PR #57, opened 2024-10-25
Modify default attachment max size and update readme to clarify #56

### LINKED ISSUE #56: Max file size could be confusing
Turns out Windows uses MB as 1024 KB not 1000 as an initial search would suggest. So I believe the current default is not actually 100MB as defined by a file system. This should be adjusted and I think it would be worth adding a note to point out this discrepancy to avoid future confusion.

joelvdavies:
## Description

See #56.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Closes #56


asuresh-code (commented):
It pretty much looks all good to me, I just have 1 question to clarify:
- is there a reason that `IMAGE__THUMBNAIL_MAX_SIZE_BYTES` was changed to `IMAGE__THUMBNAIL_MAX_SIZE_PIXELS`? Are thumbnail images 8-bit depth, so in this case they'd be equivalent?

joelvdavies:
> It pretty much looks all good to me, I just have 1 question to clarify:
> 
> * is there a reason that `IMAGE__THUMBNAIL_MAX_SIZE_BYTES` was changed to `IMAGE__THUMBNAIL_MAX_SIZE_PIXELS`? Are thumbnail images 8-bit depth, so in this case they'd be equivalent?

It was typo, I copied it `ATTACHMENT__MAX_SIZE_BYTES` and changed the first part but not the 2nd. It was always IMAGE__THUMBNAIL_MAX_SIZE_PIXELS in the env file, I just got wrong on the README.

PR #58, opened 2024-10-28
Add authentication from ims-api #6

### LINKED ISSUE #6: Add authentication from ldap-jwt-auth
- [x] Add authentication via token provided by `ldap-jwt-auth` (See `inventory-management-system-api` - these should both use the same code)
- [x] Update README.md to reflect the changes (Should be the same as `inventory-management-system-api`)

joelvdavies:
## Description

Adds authentication that functions in the same way as `inventory-management-system-api` does. The majority of this code is copied directly from there. Tested with https://github.com/ral-facilities/inventory-management-system/pull/1028.

Changes I made here have also been made on ims-api: https://github.com/ral-facilities/inventory-management-system-api/pull/408.

Also had a think about the change to using the `/verify` endpoint instead. The main issue I currently see is handling the e2e tests - we would not be able to run them in the current format without also having `ldap-jwt-auth` running/a mock endpoint as we would presumably want all requests to be denied if `/verify` is unreachable. I think this needs further investigation & discussion as it will also effect https://github.com/ral-facilities/inventory-management-system-api/issues/399. So I did not attempt to implement this here yet.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #6


VKTB:
> Also had a think about the change to using the /verify endpoint instead. The main issue I currently see is handling the e2e tests - we would not be able to run them in the current format without also having ldap-jwt-auth running/a mock endpoint as we would presumably want all requests to be denied if /verify is unreachable. I think this needs further investigation & discussion as it will also effect https://github.com/ral-facilities/inventory-management-system-api/issues/399. So I did not attempt to implement this here yet.

Good point, thanks for raising this.

VKTB (approved):

PR #61, opened 2024-10-29
Add get endpoint for image metadata

### LINKED ISSUE #36: Add an endpoint for getting a list of image metadata
Implement a GET endpoint `/images` that:
- [x] Returns a list of images
  - Should include the metadata without `object_key` 
  - Should include thumbnail
- [x] Can be given an optional query parameter `entity_id` that filters results by the `entity_id` via the database
- [x] Can be given an optional query parameter `primary` that filters results by the `primary` value via the database

asuresh-code:
## Description
Implements a `GET` `/images` that:
- [x] Returns a list of images
- [x] Accepts an optional query parameter `entity_id` that filters results by the `entity_id` via the database
- [x] Accepts an optional query parameter `primary` that filters results by the `primary` value via the database

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Review changes to test coverage

## Agile board tracking
Closes #36

### DIFF
--- a/object_storage_api/repositories/image.py
+++ b/object_storage_api/repositories/image.py
@@ -57,7 +57,7 @@ class ImageRepo:
             return ImageOut(**image)
         return None
 
-    def list(self, session: ClientSession = None) -> list[ImageOut]:
+    def list(self, entity_id: Optional[str], primary: Optional[bool], session: ClientSession = None) -> list[ImageOut]:
         """
         Retrieve Images from a MongoDB database.
 
@@ -65,5 +65,18 @@ class ImageRepo:
         :return: List of Images or an empty list if no Images are retrieved
         """
 
-        images = self._images_collection.find(session=session)
+        query = {}
+        if entity_id is not None:
+            query["entity_id":entity_id]
+        if primary is not None:
+            query["primary":primary]
+
+        message = "Retrieving all images from the database"
+        if not query:
+            logger.info(message)
+        else:
+            logger.info("%s matching the provided filter(s)", message)
+            logger.debug("Provided filter(s): %s", query)
+
+        images = self._images_collection.find(query, session=session)
         return [ImageOut(**image) for image in images]
--- a/object_storage_api/repositories/image.py
+++ b/object_storage_api/repositories/image.py
@@ -67,9 +67,9 @@ class ImageRepo:
 
         query = {}
         if entity_id is not None:
-            query["entity_id":entity_id]
+            query["entity_id"] = CustomObjectId(entity_id)
         if primary is not None:
-            query["primary":primary]
+            query["primary"] = primary
 
         message = "Retrieving all images from the database"
         if not query:
--- a/object_storage_api/repositories/image.py
+++ b/object_storage_api/repositories/image.py
@@ -56,3 +56,14 @@ class ImageRepo:
         if image:
             return ImageOut(**image)
         return None
+
+    def list(self, session: ClientSession = None) -> list[ImageOut]:
+        """
+        Retrieve Images from a MongoDB database.
+
+        :param session: PyMongo ClientSession to use for database operations.
+        :return: List of Images or an empty list if no Images are retrieved
+        """
+
+        images = self._images_collection.find(session=session)
+        return [ImageOut(**image) for image in images]
--- a/object_storage_api/repositories/image.py
+++ b/object_storage_api/repositories/image.py
@@ -68,7 +68,7 @@ class ImageRepo:
         """
 
         query = {}
-        if entity_id:
+        if entity_id is not None:
             query["entity_id"] = CustomObjectId(entity_id)
         if primary is not None:
             query["primary"] = primary
--- a/object_storage_api/repositories/image.py
+++ b/object_storage_api/repositories/image.py
@@ -59,14 +59,16 @@ class ImageRepo:
 
     def list(self, entity_id: Optional[str], primary: Optional[bool], session: ClientSession = None) -> list[ImageOut]:
         """
-        Retrieve Images from a MongoDB database.
+        Retrieve images from a MongoDB database.
 
         :param session: PyMongo ClientSession to use for database operations.
-        :return: List of Images or an empty list if no Images are retrieved
+        :param entity_id: The ID of the entity to filter images by.
+        :param primary: The primary value to filter images by.
+        :return: List of images or an empty list if no images are retrieved.
         """
 
         query = {}
-        if entity_id is not None:
+        if entity_id:
             query["entity_id"] = CustomObjectId(entity_id)
         if primary is not None:
             query["primary"] = primary
--- a/object_storage_api/routers/image.py
+++ b/object_storage_api/routers/image.py
@@ -61,10 +61,9 @@ def get_images(
     # pylint: disable=missing-function-docstring
     logger.info("Getting images")
 
-    if entity_id:
+    if entity_id is not None:
         logger.debug("Entity ID filter: '%s'", entity_id)
-
-    if primary:
+    if primary is not None:
         logger.debug("Primary filter: '%s'", primary)
 
     return image_service.list(entity_id, primary)
--- a/object_storage_api/routers/image.py
+++ b/object_storage_api/routers/image.py
@@ -49,11 +49,25 @@ def create_image(
     return image_service.create(image_metadata, upload_file)
 
 
-@router.get(path="", summary="retrieve a list of images", status_code=status.HTTP_200_OK)
+@router.get(
+    path="",
+    summary="retrieve a list of images",
+    response_description="a list of images",
+    status_code=status.HTTP_200_OK,
+)
 def get_image_list(
     image_service: ImageServiceDep,
     entity_id: Annotated[str | None, Query()] = None,
     primary: Annotated[bool | None, Query()] = None,
 ) -> list[ImageSchema]:
-    images = image_service.list()
-    return (ImageSchema(**image) for image in images)
+    # pylint: disable=missing-function-docstring
+    logger.info("Getting images")
+
+    if entity_id:
+        logger.debug("Entity ID filter: '%s'", entity_id)
+
+    if primary:
+        logger.debug("Primary filter: '%s'", primary)
+    images = image_service.list(entity_id, primary)
+
+    return (ImageSchema(**image.model_dump()) for image in images)
--- a/object_storage_api/routers/image.py
+++ b/object_storage_api/routers/image.py
@@ -6,8 +6,9 @@ service.
 import logging
 from typing import Annotated, Optional
 
-from fastapi import APIRouter, Depends, File, Form, UploadFile, status
+from fastapi import APIRouter, Depends, File, Form, Query, UploadFile, status
 
+from object_storage_api.models.image import ImageOut
 from object_storage_api.schemas.image import ImagePostMetadataSchema, ImageSchema
 from object_storage_api.services.image import ImageService
 
@@ -46,3 +47,13 @@ def create_image(
     logger.debug("Image upload file: %s", upload_file)
 
     return image_service.create(image_metadata, upload_file)
+
+
+@router.get(path="", summary="retrieve a list of images", status_code=status.HTTP_200_OK)
+def get_image_list(
+    image_service: ImageServiceDep,
+    entity_id: Annotated[str | None, Query()] = None,
+    primary: Annotated[bool | None, Query()] = None,
+) -> list[ImageSchema]:
+    images = image_service.list()
+    return (ImageSchema(**image) for image in images)
--- a/object_storage_api/routers/image.py
+++ b/object_storage_api/routers/image.py
@@ -8,7 +8,6 @@ from typing import Annotated, Optional
 
 from fastapi import APIRouter, Depends, File, Form, Query, UploadFile, status
 
-from object_storage_api.models.image import ImageOut
 from object_storage_api.schemas.image import ImagePostMetadataSchema, ImageSchema
 from object_storage_api.services.image import ImageService
 
--- a/object_storage_api/routers/image.py
+++ b/object_storage_api/routers/image.py
@@ -50,14 +50,13 @@ def create_image(
 
 @router.get(
     path="",
-    summary="retrieve a list of images",
-    response_description="a list of images",
-    status_code=status.HTTP_200_OK,
+    summary="Get images",
+    response_description="List of images",
 )
-def get_image_list(
+def get_images(
     image_service: ImageServiceDep,
-    entity_id: Annotated[str | None, Query()] = None,
-    primary: Annotated[bool | None, Query()] = None,
+    entity_id: Annotated[Optional[str], Query(description="Filter images by entity ID")] = None,
+    primary: Annotated[Optional[bool], Query(description="Filter images by primary")] = None,
 ) -> list[ImageSchema]:
     # pylint: disable=missing-function-docstring
     logger.info("Getting images")
@@ -67,6 +66,5 @@ def get_image_list(
 
     if primary:
         logger.debug("Primary filter: '%s'", primary)
-    images = image_service.list(entity_id, primary)
 
-    return (ImageSchema(**image.model_dump()) for image in images)
+    return image_service.list(entity_id, primary)
--- a/object_storage_api/services/image.py
+++ b/object_storage_api/services/image.py
@@ -11,7 +11,7 @@ from fastapi import Depends, UploadFile
 
 from object_storage_api.core.exceptions import InvalidObjectIdError
 from object_storage_api.core.image import generate_thumbnail_base64_str
-from object_storage_api.models.image import ImageIn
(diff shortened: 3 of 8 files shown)


asuresh-code:
I've been using some of the tests in the IMS API repo as a reference, but there seems to be a typo in this test; `obtained_usages_statuses` is not assigned anything, and `expected_usage_statuses` is used in its place. Can open a PR to resolve this on the api repo.
image

joelvdavies:
> obtained_usages_statuses

Yeah good spot, just make a PR for it please.

object_storage_api/routers/image.py
```diff
@@ -46,3 +46,27 @@ def create_image(
...
     return image_service.create(image_metadata, upload_file)
+
+
+@router.get(
+    path="",
+    summary="retrieve a list of images",
+    response_description="a list of images",
+    status_code=status.HTTP_200_OK,
+)
+def get_image_list(
```
joelvdavies, line 57:
```suggestion
def get_images(
```
Consistent with IMS.

object_storage_api/services/image.py
```diff
@@ -74,3 +74,12 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
         image_out = self._image_repository.create(image_in)
 
         return ImageSchema(**image_out.model_dump())
+
+    def list(self, entity_id: Optional[str], primary: Optional[bool]) -> list[ImageOut]:
+        """
+        Retrieve a list of all Images
+
```
joelvdavies, line 81: Could these also have `:param entity_id:` and `:param primary:` in the comment.

object_storage_api/repositories/image.py
```diff
@@ -56,3 +56,27 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
...
+
+        message = "Retrieving all images from the database"
+        if not query:
+            logger.info(message)
+        else:
+            logger.info("%s matching the provided filter(s)", message)
+            logger.debug("Provided filter(s): %s", query)
+
+        images = self._images_collection.find(query, session=session)
+        return [ImageOut(**image) for image in images]
```
joelvdavies, line 82: These are already ImageOut's as they come from the database, although contrary to IMS I would prefer them converted to schemas in the service layer rather than the router layer so I would suggest using this and removing the conversion in the `router.py`.
```suggestion
        return [ImageSchema(**image) for image in images]
```

object_storage_api/repositories/image.py
```diff
@@ -56,3 +56,27 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
         if image:
             return ImageOut(**image)
         return None
+
+    def list(self, entity_id: Optional[str], primary: Optional[bool], session: ClientSession = None) -> list[ImageOut]:
+        """
+        Retrieve Images from a MongoDB database.
+
+        :param session: PyMongo ClientSession to use for database operations.
+        :return: List of Images or an empty list if no Images are retrieved
```
joelvdavies, line 65:
```suggestion
        :return: List of images or an empty list if no images are retrieved.
```
Me being picky, just we changed these in IMS recently (as well as adding full stops to many comments - although mostly in the tests there).

Also the `entity_id` and `primary` are missing their comments.

test/unit/repositories/conftest.py
```diff
@@ -62,3 +64,23 @@ def mock_find_one(collection_mock: Mock, document: dict | None) -> None:
...
+        :param collection_mock: Mocked MongoDB database collection instance.
+        :param documents: The list of documents to be returned by the `find` method.
+
+        """
+        if query is None:
+            filtered_documents = documents
+        else:
+            filtered_documents = [
+                doc for doc in documents if all(query_param in doc.items() for query_param in query.items())
+            ]
```
joelvdavies, line 82: Filtering these isn't necessary here, the mock of the database query shouldn't care what query it is given. I think its worth keeping it consistent with IMS.

joelvdavies, line 82:
```suggestion
        """
```

test/unit/repositories/test_image.py
```diff
@@ -84,3 +86,79 @@ def test_create(self):
         self.mock_create(IMAGE_IN_DATA_ALL_VALUES)
         self.call_create()
         self.check_create_success()
+
+
+class ListDSL(ImageRepoDSL):
+    """Base class for `list` tests."""
+
+    _expected_image_out: list[ImageOut]
+    _obtained_image_out: list[ImageOut]
```
joelvdavies, line 95:
```suggestion
    _expected_image_out: list[ImageOut]
     _entity_id_filter: Optional[str]
     _parimary_filter: Optional[boolean]
     _obtained_image_out: list[ImageOut]
```
I would prefer this, while assigning them in the `call_list` and automatically using them in the `check_list_success`, as then the tests don't care about any implementation except the parameters the function is provided with. (IMS also does it this way)

test/unit/services/test_image.py
```diff
@@ -160,3 +161,41 @@ def test_create_with_invalid_entity_id(self):
...
+    _expected_images: MagicMock
+    _obtained_images: MagicMock
+
+    def mock_list(self) -> None:
+        """Mocks repo methods appropriately to test the `list` service method."""
+
+        self._expected_images = MagicMock()
+        self.mock_image_repository.list.return_value = self._expected_images
+
+    def call_list(self, entity_id: Optional[str] = None, primary: Optional[bool] = None) -> None:
```
joelvdavies, line 180: Missing parameters in comment.

test/unit/services/test_image.py
```diff
@@ -160,3 +161,41 @@ def test_create_with_invalid_entity_id(self):
...
+class TestList(ListDSL):
+    """Tests for listing images."""
+
+    def test_list(self):
+        """Test listing images."""
+        entity_id = str(ObjectId())
+        primary = False
+        self.mock_list()
+        self.call_list(entity_id, primary)
+        self.check_list_success()
```
joelvdavies, line 201:
```suggestion
        self.mock_list()
        self.call_list(entity_id=str(ObjectId()), primary = False)
        self.check_list_success()
```

test/e2e/test_image.py
```diff
@@ -95,3 +96,82 @@ def test_create_with_invalid_image_file(self):
...
+
+class ListDSL(CreateDSL):
+    """Base class for list tests."""
+
+    _get_response_image: Response
+
+    def get_images(self, filters: Optional[dict] = None) -> None:
+        """Gets a list of images with the given filters.
+
+        :param filters: Filters to use in the request."""
```
joelvdavies, line 109:
```suggestion
        """
        Gets a list of images with the given filters.

        :param filters: Filters to use in the request.
        """
```

test/e2e/test_image.py
```diff
@@ -95,3 +96,82 @@ def test_create_with_invalid_image_file(self):
 
         self.post_image(IMAGE_POST_METADATA_DATA_REQUIRED_VALUES_ONLY, "invalid_image.jpg")
         self.check_post_image_failed_with_detail(422, "File given is not a valid image")
+
+
+class ListDSL(CreateDSL):
```
joelvdavies, line 101:
```suggestion
# pylint:disable=fixme
# TODO: Inherit from GetDSL when added
class ListDSL(CreateDSL):
```
Just as a reminder.

test/e2e/test_image.py
```diff
@@ -95,3 +96,82 @@ def test_create_with_invalid_image_file(self):
...
+
+    def test_list_with_no_filters(self):
+        """
+        Test getting a list of all images with no filters provided.
+
+        Posts an image and expects it to be returned.
+        """
+
+        self.post_image(IMAGE_POST_METADATA_DATA_ALL_VALUES, "image.jpg")
+        expected_result = self._post_response_image.json()
```
joelvdavies, line 134: Could you add a `post_test_images` function that posts at least two images tor test the filtering on like we do for IMS e2e tests. Then you can return the actual post data and avoid having to use `self._post_response_image` directly.

test/e2e/test_image.py
```diff
@@ -95,3 +96,82 @@ def test_create_with_invalid_image_file(self):
...
+        self.get_images(filters={"primary": False})
+        self.check_get_images_success([expected_result])
+
+    def test_list_with_primary_filter_with_no_matching_results(self):
+        """Test getting a list of all images with a `primary` filter
+
+        Posts nothing and expects no results."""
+        self.post_image(IMAGE_POST_METADATA_DATA_ALL_VALUES, "image.jpg")
+        self.get_images(filters={"primary": True})
+        self.check_get_images_success([])
```
joelvdavies, line 177: I think we also need tests for invalid `entity_id` and `primary` values.

joelvdavies (changes requested):
Works well, it worked for Joshua on the frontend too 🥳. Mostly minor comments to increase consistency with IMS.

object_storage_api/repositories/image.py
asuresh-code: Just double checking, if I am converting them to schemas in the service layer, would I not be returning as them `ImageOut` in the Repo layer like so

`return [ image for image in images]`
instead of 
`return [ImageSchema(**image) for image in images]`

object_storage_api/repositories/image.py
asuresh-code: Is the text in comments meant to be normal grammar (i.e no mid-sentence capitalisation unless a proper noun, acronym, etc)?

In that case should `Retrieve Images from a MongoDB database.` , be changed to `Retrieve images from a MongoDB database.`>

joelvdavies: Yes like that. When I added systems on IMS all my comments referred to them as `System`, when refactoring all of the tests I realised Viktor's style always used the lower case with full stops so I matched his and used the lower case variant instead `system`. As I said it was a picky point to match what happened on IMS. (Not that the code there actually got changed, just the tests)

test/unit/repositories/test_image.py
```diff
@@ -84,3 +86,84 @@ def test_create(self):
...
+        self._obtained_image_out = self.image_repository.list(
+            session=self.mock_session, entity_id=entity_id, primary=primary
+        )
+
+    def check_list_success(self) -> None:
+        """Checks that a prior call to `call_list` worked as expected."""
+        expected_query = {}
+        if self._entity_id_filter:
+            expected_query["entity_id"] = ObjectId(self._entity_id_filter)
+        if self._primary_filter != None:
```
joelvdavies, line 131:
```suggestion
        if self._entity_id_filter is not None:
            expected_query["entity_id"] = ObjectId(self._entity_id_filter)
        if self._primary_filter is not None:
```
Resolves pylint issue
image

test/unit/repositories/test_image.py
```diff
@@ -84,3 +86,84 @@ def test_create(self):
...
+    def test_list_with_primary(self):
+        """Test listing all images with a primary argument."""
+        self.mock_list([IMAGE_IN_DATA_ALL_VALUES])
+        self.call_list(primary=False)
+        self.check_list_success()
+
+    def test_list_with_primary_and_entity_id(self):
+        """Test listing all images with an entity_id and primary argument."""
+        self.mock_list([IMAGE_IN_DATA_ALL_VALUES])
+        self.call_list(primary=False, entity_id=IMAGE_IN_DATA_ALL_VALUES["entity_id"])
```
joelvdavies, line 168:
```suggestion
        self.call_list(primary=True, entity_id=IMAGE_IN_DATA_ALL_VALUES["entity_id"])
```
Then both cases are covered.

test/unit/services/test_image.py
```diff
@@ -160,3 +161,48 @@ def test_create_with_invalid_entity_id(self):
...
+    def mock_list(self, image_in_data: list[dict]) -> None:
+        """Mocks repo methods appropriately to test the `list` service method."""
+
+        self._expected_images = [
+            ImageSchema(**ImageOut(**ImageIn(**image_in_data).model_dump()).model_dump())
+            for image_in_data in image_in_data
+        ]
+        self.mock_image_repository.list.return_value = [
+            ImageOut(**ImageIn(**image_in_data).model_dump()) for image_in_data in image_in_data
+        ]
```
joelvdavies, line 183:
```suggestion
    def mock_list(self) -> None:
        """Mocks repo methods appropriately to test the `list` service method."""

        # Just returns the result after converting it to the schemas currently, so actual value doesn't matter here
        images_out = [ImageOut(**ImageIn(**IMAGE_IN_DATA_ALL_VALUES).model_dump())]
        self.mock_image_repository.list.return_value = images_out
        self._expected_images = [ImageSchema(**image_out.model_dump()) for image_out in images_out]
```
I would just do something like this, the actual value doesnt matter the same as the MagicMocks on IMS. As long as the data is valid it will test the result is returned after going through the schema. Then the test function doesn't even need to provide this.

test/e2e/test_image.py
```diff
@@ -95,3 +96,177 @@ def test_create_with_invalid_image_file(self):
...
+            be required for an `ImageSchema`.
+        """
+        assert self._get_response_image.status_code == 200
+        assert self._get_response_image.json() == expected_images_get_data
+
+    def check_images_list_response_failed_with_message(self, status_code, expected_detail, obtained_detail):
+        """Checks the response of listing images failed as expected"""
+
+        assert self._get_response_image.status_code == status_code
+        assert obtained_detail == expected_detail
```
joelvdavies, line 130: Could these go below the post_test_images so the order is consistent with IMS please.

test/e2e/test_image.py
```diff
@@ -95,3 +96,177 @@ def test_create_with_invalid_image_file(self):
...
+        image_a_id = self.post_image({**IMAGE_POST_METADATA_DATA_ALL_VALUES, "entity_id": entity_id_a}, "image.jpg")
+
+        # Second item
+        image_b_id = self.post_image(
+            {
+                **IMAGE_POST_METADATA_DATA_ALL_VALUES,
+                "entity_id": entity_id_a,
+                "title": "Essay Title",
+                "description": "A research essay.",
+            },
```
joelvdavies, line 152:
```suggestion
            {
                **IMAGE_POST_METADATA_DATA_REQUIRED_VALUES_ONLY,
                "entity_id": entity_id_a,
            },
```
Its nicer to use data from the mock data where possible to avoid repeatedly defining fields. The title and description doesn't really need to be different for each item with a single one we are already checking it gets used correctly as there is no specific logic to avoid duplicates etc, so C can probably be identical to A or B except the entity ID. Should mate the return values have less fields as well.

test/e2e/test_image.py
```diff
@@ -95,3 +96,177 @@ def test_create_with_invalid_image_file(self):
...
+
+
+class TestList(ListDSL):
+    """Tests for getting a list of images."""
+
+    def test_list_with_no_filters(self):
+        """
+        Test getting a list of all images with no filters provided.
+
+        Posts 3 images and expects it to be returned.
```
joelvdavies, line 199:
```suggestion
        Posts 3 images and expects all of them to be returned.
```

test/e2e/test_image.py
```diff
@@ -95,3 +96,177 @@ def test_create_with_invalid_image_file(self):
...
+    def test_list_with_entity_id_filter(self):
+        """
+        Test getting a list of all images with an `entity_id` filter provided.
+
+        Posts 3 images and then filter using the `entity_id`.
+        """
+
+        expected_result = self.post_test_images()[:2]
+        self.get_images(filters={"entity_id": expected_result[0]["entity_id"]})
+        self.check_get_images_success(expected_result)
```
joelvdavies, line 215:
```suggestion
        images = self.post_test_images()
        self.get_images(filters={"entity_id": expected_result[0]["entity_id"]})
        self.check_get_images_success(images[0])
```
Just my preference, I think its slightly easier to read.

test/e2e/test_image.py
```diff
@@ -95,3 +96,177 @@ def test_create_with_invalid_image_file(self):
...
+        """
+        self.post_test_images()
+        self.get_images(filters={"primary": str(ObjectId())})
+        self.check_images_list_response_failed_with_message(
+            422,
+            "Input should be a valid boolean, unable to interpret input",
+            self._get_response_image.json()["detail"][0]["msg"],
+        )
+
+    def test_list_with_invalid_entity_id_filter(self):
```
joelvdavies, line 262: Could this test be moved to be at the end of the entity_id tests so they are grouped.

test/e2e/test_image.py
asuresh-code: So should `expected_result` be renamed to `images` in all of the tests in this file, or just in this one? Also would have to look more like this, since 1. expected_result is not defined and 2. the expected result is the 1st 2 images posted, not just the 1st one
```suggestion
        images = self.post_test_images()
        self.get_images(filters={"entity_id": images[0]["entity_id"]})
        self.check_get_images_success(images[:2])
```

joelvdavies (approved):
LGTM 🎉

VKTB (approved):
Good work!

Issue #15, opened 2024-09-25
Investigate improved error handling


joelvdavies:
Essentially testing out https://github.com/ral-facilities/inventory-management-system-api/issues/74. #12 uses this for invalid object id errors which I think is one of the trickier ones to use this method for. Undecided my self if its actually better, this issue is essentially a note to re-evaluate it later and either revert it to the IMS method (catching all errors in the router) or create an issue to apply what is done here to `inventory-management-system-api`.

joelvdavies:
I think it would be worth looking into https://github.com/ral-facilities/object-storage-api/pull/69#discussion_r1867959410 as well.

PR #69, opened 2024-11-21
Add Get Endpoint for Image metadata and presigned url

### LINKED ISSUE #37: Add an endpoint for getting an image with a download URL
Implement a GET endpoint `/images/<image_id>` that:
- [ ] Generates a pre-signed download URL for the image file
  - Should use the original `file_name` for the download? - Ideally would open automatically on a browser due to using an image content type.
- [ ] Returns the image metadata and thumbnail
    - Excluding `object_key`
    - Including the download url

asuresh-code:
## Description
See #37.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #37


asuresh-code:
One of the requirements for this issue listed here https://github.com/ral-facilities/object-storage-api/issues/37#issue-2570131775 , requires that the endpoint:

> Must gracefully deal with a situation when the file object with the given object_key doesn't exist

When given an object key that doesn't exist, the presigned url is still generated without any error being thrown. The issue only arises when you open the url and are greeted with this response.

image

I wanted to check if this would still be a requirement for the back-end to handle or the front-end? S3 won't throw any exceptions when generating the url, so if the check has to be done in the back-end, it would require an additional call to S3 to get the file, implemeting something like this https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3

joelvdavies:
> One of the requirements for this issue listed here [#37 (comment)](https://github.com/ral-facilities/object-storage-api/issues/37#issue-2570131775) , requires that the endpoint:
> 
> > Must gracefully deal with a situation when the file object with the given object_key doesn't exist
> 
> When given an object key that doesn't exist, the presigned url is still generated without any error being thrown. The issue only arises when you open the url and are greeted with this response.
> 
> image
> 
> I wanted to check if this would still be a requirement for the back-end to handle or the front-end? S3 won't throw any exceptions when generating the url, so if the check has to be done in the back-end, it would require an additional call to S3 to get the file, implemeting something like this https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3
> 
> A similar questions was asked here https://stackoverflow.com/questions/78010875/how-to-directly-generate-presigned-url-if-key-exist-in-s3-or-upload-obj-and-cre, and the response was to add an explicit check referenced in my 1st link

Yeah this behaviour is fine. I just checked the front end and it displays the right message when it fails to load from the URL so I it's fine to leave the URL in and ignore the check.

object_storage_api/stores/image.py
```diff
@@ -37,3 +38,22 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
         )
 
         return object_key
+
+    def create_presigned_url(self, object_key: str, file_name: str) -> HttpUrl:
```
joelvdavies, line 42:
```suggestion
    def create_presigned_get(self, object_key: str, file_name: str) -> HttpUrl:
```
Would match with attachments `create_presigned_post` as this is a get presigned URL.

object_storage_api/stores/image.py
```diff
@@ -37,3 +38,22 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
...
+        """Generate a presigned URL to share an S3 object.
+
+        :param object_key: object key of the image to generate the url for.
+        :return: Image Metadata with a presigned url.
```
joelvdavies, line 46: Missing `param: file_name`. You could even just take in the `ImageOut` instead and let it use the parameters directly. For attachments it takes in the attachment post object and returns the object key as I wanted to leave all logic on what to use for the presigned url inside the function itself.

object_storage_api/services/image.py
```diff
@@ -75,6 +75,25 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
 
         return ImageSchema(**image_out.model_dump())
 
+    def get(self, image_id: str) -> ImageGetUrlInfoSchema:
+        """
+        Retrieve an image with its presigned url by its ID.
+        :param: ID of the image to retrieve.
+        :return: An image or None if no image is retrieved.
```
joelvdavies, line 82:
```suggestion
        Retrieve an image with its presigned url by its ID.
        
        :param: ID of the image to retrieve.
        :return: An image or `None` if no image is retrieved.
```
You should also add `:raises` here like in the create method above. Doesn't return `None` anymore.

object_storage_api/services/image.py
```diff
@@ -75,6 +75,25 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
...
+            image = self._image_repository.get(image_id=image_id)
+            if not image:
+                raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message)
+            presigned_url = self._image_store.create_presigned_url(
+                object_key=image.object_key, file_name=image.file_name
+            )
+            return ImageGetUrlInfoSchema(**image.model_dump(), url=presigned_url)
+        except InvalidObjectIdError as exc:
+            logger.exception("The ID is not a valid ObjectId value")
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
```
joelvdavies, line 95: I would create a new exception for this rather than raising a HTTPException in the service as these should be reserved to the router as services could in theory be run independently of the actual API. If we followed ims-api then this wouldn't error here and the response would be handled in the router. Otherwise `MissingRecordError` from IMS would work but perhaps that would make more sense if it was raised inside the repo layer. So maybe `MissingEntityError` if we keep it in the service.

test/e2e/test_image.py
```diff
@@ -100,11 +100,65 @@ def test_create_with_invalid_image_file(self):
 
 # pylint:disable=fixme
 # TODO: Inherit from GetDSL when added
```
joelvdavies, line 102: You can remove this now.

object_storage_api/routers/image.py
```diff
@@ -67,3 +68,20 @@ def get_images(
...
+    logger.info("Getting image with ID %s", image_id)
+    message = "An image with such ID was not found"
+    try:
+        image = image_service.get(image_id)
+        if not image:
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message)
+        return ImageGetUrlInfoSchema(**image.model_dump())
+    except InvalidObjectIdError as exc:
+        logger.exception("The ID is not a valid ObjectId value")
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
```
joelvdavies, line 87: This doesn't need to be repeated due to the exception middleware already detecting any exceptions inheriting from `BaseAPIException`. Same for the one above depending on what we decide to do about the error in the service layer.

joelvdavies (commented):
Some initial comments, one I am not sure on but we can discuss, worked well on the front end when we tested it.

object_storage_api/repositories/image.py
```diff
@@ -49,13 +50,21 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
...
-        image_id = CustomObjectId(image_id)
         logger.info("Retrieving image with ID: %s from the database", image_id)
-        image = self._images_collection.find_one({"_id": image_id}, session=session)
+        try:
+            image_id = CustomObjectId(image_id)
+            image = self._images_collection.find_one({"_id": image_id}, session=session)
+        except InvalidObjectIdError as e:
+            exc = MissingRecordError(f"The image_id {image_id} is not valid.")
+            exc.response_detail = f"Invalid image_id format: {image_id}"
+            raise exc from e
```
joelvdavies, line 62:
```suggestion
        except InvalidObjectIdError as e:
            exc = MissingRecordError(f"The image_id {image_id} is not valid.")
            exc.response_detail = "Image not found"
            raise exc from e
```
I would go with `Image not found` as this is like Viktor requested, thanks to the raise exc from e - it logs the earlier one too specifying its due to an invalid id.

I would also modify BaseAPIException's init to take in an optional response detail that assigns it if it isn't None, so it can be put on one line e.g. `raise MissingRecordError(detail, response_detail) from exc`. You could also override the init in `MissingRecordError` to take in the `entity_name` and automatically assigning the response detail to `f"{entity_name.capitalize()} was not found"` to make it even simpler.

object_storage_api/repositories/image.py
```diff
@@ -49,13 +50,21 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
...
         if image:
             return ImageOut(**image)
-        return None
+        exc = MissingRecordError(f"Image with image_id {image_id} was not found.")
+        exc.response_detail = f"Image with image_id {image_id} was not found."
+        raise exc
```
joelvdavies, line 67: You can do what we do in IMS and have
```message = f"Image with image_id {image_id} was not found."```
above and then use it in both exceptions. Same for the response detail.

File: object_storage_api/routers/image.py [hunk 9]
```diff
@@ -67,3 +67,13 @@ def get_images(
         logger.debug("Primary filter: '%s'", primary)
 
     return image_service.list(entity_id, primary)
+
+
+@router.get(path="/{image_id}", summary="Get an image by ID", response_description="Single image")
+def get_image(
+    image_id: Annotated[str, Path(description="The ID of the item to get")], image_service: ImageServiceDep
```
joelvdavies, line 74:
```suggestion
    image_id: Annotated[str, Path(description="ID of the image to get")], image_service: ImageServiceDep
```

object_storage_api/repositories/image.py
```diff
@@ -49,13 +50,21 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
         :param image_id: ID of the image to retrieve.
         :param session: PyMongo ClientSession to use for database operations.
         :return: Retrieved image or `None` if not found.
+        :raises MissingRecordError: If the supplied image_id is invalid or non-existent.
```
joelvdavies, line 53:
```suggestion
        :raises MissingRecordError: If the supplied `image_id` is invalid or non-existent.
```

object_storage_api/services/image.py
```diff
@@ -75,6 +75,16 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
 
         return ImageSchema(**image_out.model_dump())
 
+    def get(self, image_id: str) -> ImageGetUrlInfoSchema:
+        """
+        Retrieve an image with its presigned url by its ID.
+        :param: ID of the image to retrieve.
+        :return: An image or None if no image is retrieved.
```
joelvdavies, line 82:
```suggestion
        Retrieve an image with its presigned url by its ID.
        
        :param: ID of the image to retrieve.
        :return: An image or None if no image is retrieved.
```

object_storage_api/stores/image.py
```diff
@@ -37,3 +39,22 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
         )
 
         return object_key
+
+    def create_presigned_get(self, image: ImageOut) -> HttpUrl:
+        """Generate a presigned URL to share an S3 object.
+
+        :param image: `ImageOut` model of the image.
+        :return: Image metadata with a presigned url.
+        """
```
joelvdavies, line 48:
```suggestion
    def create_presigned_get(self, image: ImageOut) -> str:
        """
        Generate a presigned URL to get an S3 object.

        :param image: `ImageOut` model of the image.
        :return: Image metadata with a presigned url.
        """
```
I tried printing out the type(response) and it was a str instead of HttpUrl.

test/unit/stores/test_image.py
```diff
@@ -82,3 +84,54 @@ def test_upload(self):
         self.mock_upload(IMAGE_POST_METADATA_DATA_ALL_VALUES)
         self.call_upload()
         self.check_upload_success()
+
+
+class CreatePresignedURLDSL(ImageStoreDSL):
+    """Base class for `create` tests."""
+
+    _image: ImageOut
```
joelvdavies, line 92:
```suggestion
    _image_out: ImageOut
```

test/unit/stores/test_image.py
```diff
@@ -82,3 +84,54 @@ def test_upload(self):
...
+    _expected_presigned_url: HttpUrl
+    _obtained_presigned_url: HttpUrl
+
+    def mock_create_presigned_get(self) -> None:
```
joelvdavies, line 96:
```suggestion
    def mock_create_presigned_get(self, image_in_data: dict) -> None:
```
I would handle this like the repo and take in the data that it will test with and use that to construct the image out model so future tests could change it.

test/unit/services/test_image.py
```diff
@@ -163,6 +163,45 @@ def test_create_with_invalid_entity_id(self):
         self.check_create_failed_with_exception("Invalid ObjectId value 'invalid-id'")
 
 
+class GetDSL(ImageServiceDSL):
+    """Base class for `get` tests."""
+
+    _image_id: str
```
joelvdavies, line 169:
```suggestion
    __obtained_image_id: str
```
To match IMS.

test/unit/services/test_image.py
```diff
@@ -163,6 +163,45 @@ def test_create_with_invalid_entity_id(self):
...
+        """Calls the `ImageService` `get` method.
+
+        :param image_id: The ID of the image to retrieve.
+        """
+        self._image_id = image_id
+        self._obtained_image = self.image_service.get(image_id=image_id)
+
+    def check_get_success(self) -> None:
+        """Checks that a prior call to `call_get` worked as expected."""
+        self.mock_image_repository.get.assert_called_once_with(image_id=self._image_id)
```
joelvdavies, line 191: I think this should also assert `create_presigned_get` was called as well.

test/unit/repositories/test_image.py
```diff
@@ -88,6 +88,50 @@ def test_create(self):
         self.check_create_success()
 
 
+class GetDSL(ImageRepoDSL):
+    """Base class for `get` tests."""
+
+    _image_id: str
```
joelvdavies, line 94:
```suggestion
    _obtained_image_id: str
```

test/unit/repositories/test_image.py
```diff
@@ -88,6 +88,50 @@ def test_create(self):
...
+
+        RepositoryTestHelpers.mock_find_one(self.images_collection, self._expected_image_out.model_dump())
+
+    def call_get(self, image_id: str) -> None:
+        """Calls the `ImageRepo` `get method` method.
+
+        :param image_id: The ID of the image to retrieve.
+        """
+        self._image_id = image_id
+        self._obtained_image_out = self.image_repository.get(session=self.mock_session, image_id=image_id)
```
joelvdavies, line 115:
```suggestion
        self._obtained_image_out = self.image_repository.get(image_id=image_id, session=self.mock_session)
```

test/unit/repositories/test_image.py
```diff
@@ -88,6 +88,50 @@ def test_create(self):
...
+
+    def check_get_success(self) -> None:
+        """Checks that a prior call to `call_get` worked as expected."""
+        expected_query = {"_id": ObjectId(self._image_id)}
+
+        self.images_collection.find_one.assert_called_once_with(expected_query, session=self.mock_session)
```
joelvdavies, line 121:
```suggestion

        self.images_collection.find_one.assert_called_once_with("_id": ObjectId(self._image_id)}, session=self.mock_session)
```

test/unit/repositories/test_image.py
```diff
@@ -88,6 +88,50 @@ def test_create(self):
...
+    _image_id: str
+    _expected_image_out: ImageOut
+    _obtained_image_out: ImageOut
+
+    def mock_get(self, image_in_data: dict) -> None:
+        """
+        Mocks database methods appropriately to test the `get` repo method.
+
+        :param image_in_data: Dictionary containing the image data as would be required for an
+            `ImageIn` database model (i.e. no ID or created and modified times required).
```
joelvdavies, line 103:
```suggestion
            `ImageIn` database model (i.e. no created and modified times required).
```
In this case they do have the ID.

test/unit/repositories/test_image.py
```diff
@@ -88,6 +88,50 @@ def test_create(self):
...
+
+    def check_get_success(self) -> None:
+        """Checks that a prior call to `call_get` worked as expected."""
+        expected_query = {"_id": ObjectId(self._image_id)}
+
+        self.images_collection.find_one.assert_called_once_with(expected_query, session=self.mock_session)
+        assert self._obtained_image_out == self._expected_image_out
+
+
+class TestGet(GetDSL):
```
joelvdavies, line 125: As the repo now raises the errors, this should have the tests for it like the SystemRepo on IMS does.

object_storage_api/schemas/image.py
```diff
@@ -28,3 +28,11 @@ class ImageSchema(CreatedModifiedSchemaMixin, ImagePostMetadataSchema):
     file_name: str = Field(description="File name of the image")
     primary: bool = Field(description="Whether the image is the primary for its related entity")
     thumbnail_base64: str = Field(description="Thumbnail of the image as a base64 encoded byte string")
+
+
+class ImageGetUrlInfoSchema(ImageSchema):
```
joelvdavies, line 33: I think given the way we have named things in IMS `ImageGetUrlInfoSchema` would imply it is the UrlInfo object inside the ImageGetSchema. I think the previous ImageSchema should be renamed ImageMetadataSchema, so that this can be the true ImageGet.

test/e2e/test_image.py
```diff
@@ -98,13 +98,64 @@ def test_create_with_invalid_image_file(self):
...
+        """
+        Checks that a prior call to `get_image` gave a successful response with the expected data returned.
+
+        :param expected_image_data: Dictionary containing the expected image data as would be required
+            for an `ImageSchema`.
+        """
+        assert self._get_response_image.status_code == 200
+        response_json = self._get_response_image.json()
+        del response_json["url"]
+        assert response_json == expected_image_data
```
joelvdavies, line 124: If renaming the schemas the IMAGE_GET_DATA_ALL_VALUES an become IMAGE_METDATA_GET_DATA_ALL_VALUES, then you can add a new one
```IMAGE_GET_DATA_ALL_VALUES = {**IMAGE_METDATA_GET_DATA_ALL_VALUES, url: ANY}```
Allowing this to be
```suggestion
        assert self._get_response_image.json() == expected_image_data
```

joelvdavies (commented):
Just some minor comments and suggestions.

test/unit/services/test_image.py
asuresh-code: Does it make sense for it to be called `_obtained...`? 

Because in the test it's only being used to make-up the query, it's not a result from either the `call `or `mock` functions. It is used in the `check_success` functions to assert that the request was made properly, so if it had to have one of the prefixes, it would arguably be `_expected`?

test/unit/repositories/test_image.py
asuresh-code: Same question mentioned [here](https://github.com/ral-facilities/object-storage-api/pull/69#discussion_r1860853656)

test/unit/services/test_image.py
joelvdavies: I was thinking of it like
_obtained_image_id - Id of the image being obtained via the get
_expected_image - Expected image returned by the get
_obtained_image - Actual obtained image returned by the get

Its not really the expected one as its the one passed into the 'get' function in the first place.

object_storage_api/repositories/image.py
```diff
@@ -49,13 +50,21 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
...
         logger.info("Retrieving image with ID: %s from the database", image_id)
-        image = self._images_collection.find_one({"_id": image_id}, session=session)
+        message = f"Image with image_id {image_id} was not found."
+        entity_name = "image"
+        try:
+            image_id = CustomObjectId(image_id)
+            image = self._images_collection.find_one({"_id": image_id}, session=session)
+        except InvalidObjectIdError as e:
+            exc = MissingRecordError(detail=message, entity_name=entity_name)
+            raise exc from e
```
joelvdavies, line 63:
```suggestion
        except InvalidObjectIdError as exc:
            raise MissingRecordError(detail=message, entity_name=entity_name) from exc
```
This and the one below can go on one line now.

test/unit/repositories/test_image.py
```diff
@@ -88,6 +89,108 @@ def test_create(self):
...
+
+    def check_get_success(self) -> None:
+        """Checks that a prior call to `call_get` worked as expected."""
+
+        self.images_collection.find_one.assert_called_once_with(
+            {"_id": ObjectId(self._obtained_image_id)}, session=self.mock_session
+        )
+        assert self._obtained_image_out == self._expected_image_out
+
+    def check_get_failed_with_exception(self, image_id: str, assert_find: bool) -> None:
```
joelvdavies, line 145:
```suggestion
    def check_get_failed_with_exception(self, assert_find: bool) -> None:
```
Can you just use `_obtained_image_id` here?

test/unit/stores/test_image.py
```diff
@@ -82,3 +84,56 @@ def test_upload(self):
...
+        assert self._obtained_presigned_url == self._expected_presigned_url
+
+
+class TestCreatePresignedURL(CreatePresignedURLDSL):
+    """Tests for creating a presigned url for an image."""
+
+    def test_create_presigned_post(self):
+        """Test creating a presigned url for an image."""
+
+        self.mock_create_presigned_get(image_in_data=IMAGE_IN_DATA_ALL_VALUES)
```
joelvdavies, line 137:
```suggestion
        self.mock_create_presigned_get(IMAGE_IN_DATA_ALL_VALUES)
```

joelvdavies (commented):
Looks good, final few minor comments. Will get Viktor to have a look over the error handling next week to see what he thinks.

object_storage_api/repositories/image.py
```diff
@@ -49,13 +50,19 @@ def get(self, image_id: str, session: ClientSession = None) -> Optional[ImageOut
...
-        image_id = CustomObjectId(image_id)
         logger.info("Retrieving image with ID: %s from the database", image_id)
-        image = self._images_collection.find_one({"_id": image_id}, session=session)
+        message = f"Image with image_id {image_id} was not found."
+        entity_name = "image"
+        try:
+            image_id = CustomObjectId(image_id)
+            image = self._images_collection.find_one({"_id": image_id}, session=session)
+        except InvalidObjectIdError as exc:
+            raise MissingRecordError(detail=message, entity_name=entity_name) from exc
```
joelvdavies, line 62: Sorry just had another think about he errors and where they should be raised, I think I have changed my mind from before 😄

Thinking about the repo as the interface to the database – I would expect that doing `repo.get(“invalid-id”)` should raise an invalid id error. As an example, for IMS within `CatalogueCategoryService.create`’s parent check, we use the `repo.get` – if the parent id was actually invalid but we checked existence by catching the MissingRecordError then creating a catalogue category with a `parent_id: “invalid-id”` would succeed with a `parent_id` of `None` when it should fail due to an invalid ID. 

Due to this I actually think that this part needs to be in the service instead 🤦‍♂️

asuresh-code: To differentiate between an invalid ID and missing record we could look at the `detail `of the `MissingRecordError` in the service layer. Implementing something like this, whilst keeping most of the error handling in the repo layer.

```
parent_id = catalogue_category.parent_id
        try:
            parent_catalogue_category = self.get(parent_id) if parent_id else None
        except MissingRecordError as exc:
            if exc.detail == f"Invalid catalogue category ID: {parent_id}":
                raise exc
            elif exc.detail == f"Missing catalogue category: {parent_id}":
                
```

But I do agree, that from a "layer's purpose" perspective, the repo layer should just be throwing errors related to the database, whilst the service layer takes care of more of the "business logic".

asuresh-code: Actually, I've been thinking about it more and I have another idea.

In the repo layer, we could raise a `MissingRecordError` and `InvalidObjectIDerror` respectively for the 2 separate error cases. We could change the `status_code `of the `InvalidObjectIDerror` when its caught in the repo layer to a `404`, and let it raise as normal. In the service layer this makes it easy to differentiate it, and when the response reaches the user, it just seems like the entity didn't exist since the status code and response detail is set to `404 `and `Entity not found` respectively.

joelvdavies: That sounds perfect, the value logged would still show it was due to the invalid id but when raised it would still look the same. Good idea!

object_storage_api/repositories/image.py
```diff
@@ -42,20 +43,27 @@ def create(self, image: ImageIn, session: ClientSession = None) -> ImageOut:
...
         """
-        image_id = CustomObjectId(image_id)
         logger.info("Retrieving image with ID: %s from the database", image_id)
-        image = self._images_collection.find_one({"_id": image_id}, session=session)
+        try:
+            image_id = CustomObjectId(image_id)
+            image = self._images_collection.find_one({"_id": image_id}, session=session)
+        except InvalidObjectIdError as exc:
+            exc.status_code = 404
+            exc.response_detail = "Image not found"
```
VKTB, line 62: What are the cases where `InvalidObjectIdError` has a different `status_code` and `response_detail`?

object_storage_api/repositories/image.py
```diff
@@ -42,20 +43,27 @@ def create(self, image: ImageIn, session: ClientSession = None) -> ImageOut:
...
+            raise exc
         if image:
             return ImageOut(**image)
-        return None
+        raise MissingRecordError(detail=f"Image with image_id {image_id} was not found.", entity_name="image")
```
VKTB, line 66: It might be nice to keep the message similar to the format used in IMS.
```suggestion
        raise MissingRecordError(detail=f"No image found with ID: {image_id}", entity_name="image")
```

object_storage_api/routers/image.py
```diff
@@ -67,3 +67,13 @@ def get_images(
...
 
     return image_service.list(entity_id, primary)
+
+
+@router.get(path="/{image_id}", summary="Get an image by ID", response_description="Single image")
+def get_image(
+    image_id: Annotated[str, Path(description="ID of the image to get")], image_service: ImageServiceDep
+) -> ImageSchema:
+    # pylint: disable=missing-function-docstring
+    logger.info("Getting image with ID %s", image_id)
```
VKTB, line 77:
```suggestion
    logger.info("Getting image with ID: %s", image_id)
```

object_storage_api/repositories/image.py
joelvdavies: @VKTB The issue here is that the `InvalidObjectIdError` is raised when any schema fails to validate due to an invalid ID, normally when this happens we expect it to be a 422 as it happens in request bodies, which is why its the default when raised in `CustomObjectId`. Here we are catching it just because its a specific case for the get request where we are trying to make it return a 404 as its actually an endpoint path parameter instead. So I think its worse having it as  404, as all other requests would have to put it back to a 422 again vs one get having to do it.

joelvdavies: I meant something like
```python
def check_path_id(v: str) -> str:
    try
        CustomObjectId(v)
    except InvalidIdError as exc
         exc.status_code = 404
         raise exc

PathIdStr = Annotated[str, AfterValidator[check_path_id]]
```
Then all router endpoints that has an id in its path param use `PathIdStr` instead of str, so only those will validate the id in this way and the rest presumably happening in schemas are 422.

VKTB: I see what you meant. That could probably work, yes but I will leave it to you guys to decide if you want to change it especially if the current approach already works.

joelvdavies: Yeah it would need an entity name as well anyway so the responses matched, so it could also be preferable to keep it as is so they are both in the same place. I would probably just leave it for this PR.

asuresh-code: I've just pushed a commit where the validator accepts the entity name as well, at the router level.

Can roll it back to being raised at the repo level though, if we want to leave it for a separate issue.

joelvdavies: Yeah I would prefer it to be addressed separately, I think there could be a few ways to consider e.g. https://docs.pydantic.dev/latest/concepts/validation_decorator/#extension-validating-arguments-before-calling-a-function and we need this PR merged to unblock reviewing some front end ones. I will add it to https://github.com/ral-facilities/object-storage-api/issues/15.

joelvdavies (dismissed):

test/unit/repositories/test_image.py
```diff
@@ -88,6 +89,108 @@ def test_create(self):
...
+        if image_in_data:
+            image_in_data["id"] = image_id
+        self._expected_image_out = ImageOut(**ImageIn(**image_in_data).model_dump()) if image_in_data else None
+
+        RepositoryTestHelpers.mock_find_one(
+            self.images_collection, self._expected_image_out.model_dump() if self._expected_image_out else None
+        )
+
+    def call_get(self, image_id: str) -> None:
+        """Calls the `ImageRepo` `get` method.
```
rowan04, line 117:
```suggestion
        """
        Calls the `ImageRepo` `get` method.
```

test/unit/services/test_image.py
```diff
@@ -163,21 +163,64 @@ def test_create_with_invalid_entity_id(self):
...
+
+        self._expected_image_out = ImageOut(**ImageIn(**IMAGE_IN_DATA_ALL_VALUES).model_dump())
+        self.mock_image_repository.get.return_value = self._expected_image_out
+        self.mock_image_store.create_presigned_get.return_value = "https://fakepresignedurl.co.uk"
+        self._expected_image = ImageSchema(
+            **self._expected_image_out.model_dump(), url="https://fakepresignedurl.co.uk"
+        )
+
+    def call_get(self, image_id: str) -> None:
+        """Calls the `ImageService` `get` method.
```
rowan04, line 185:
```suggestion
        """
        Calls the `ImageService` `get` method.
```

rowan04 (commented):
Minor docstring changes?

object_storage_api/core/exceptions.py
```diff
@@ -53,3 +60,26 @@ class InvalidImageFileError(BaseAPIException):
 
     status_code = 422
     response_detail = "File given is not a valid image"
+
+
+class MissingRecordError(DatabaseError):
+    """
+    A specific database record was requested but could not be found.
+    """
```
rowan04, line 68:
```suggestion
    """A specific database record was requested but could not be found."""
```

object_storage_api/schemas/image.py
```diff
@@ -19,12 +19,20 @@ class ImagePostMetadataSchema(BaseModel):
...
     id: str = Field(description="ID of the image")
     file_name: str = Field(description="File name of the image")
     primary: bool = Field(description="Whether the image is the primary for its related entity")
     thumbnail_base64: str = Field(description="Thumbnail of the image as a base64 encoded byte string")
+
+
+class ImageSchema(ImageMetadataSchema):
+    """
+    Schema model for an image get request response.
+    """
```
rowan04, line 36:
```suggestion
    """Schema model for an image get request response."""
```

object_storage_api/schemas/image.py
```diff
@@ -19,12 +19,20 @@ class ImagePostMetadataSchema(BaseModel):
     description: Optional[str] = Field(default=None, description="Description of the image")
 
 
-class ImageSchema(CreatedModifiedSchemaMixin, ImagePostMetadataSchema):
+class ImageMetadataSchema(CreatedModifiedSchemaMixin, ImagePostMetadataSchema):
     """
-    Schema model for an image get request response.
+    Schema model for an image's metadata.
     """
```
rowan04, line 25: This could be made into a single line docstring too. It won't let me write the change myself as it has a deleted line in it.

object_storage_api/stores/image.py
```diff
@@ -37,3 +38,22 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
         )
 
         return object_key
+
+    def create_presigned_get(self, image: ImageOut) -> str:
+        """Generate a presigned URL to share an S3 object.
```
rowan04, line 43:
```suggestion
        """
        Generate a presigned URL to share an S3 object.
```

test/unit/stores/test_image.py
```diff
@@ -82,3 +83,56 @@ def test_upload(self):
...
+        """
+        self._image_out = ImageOut(**ImageIn(**image_in_data).model_dump())
+
+        # Mock presigned url generation
+        self._expected_presigned_url = "example_presigned_url"
+        self.mock_s3_client.generate_presigned_url.return_value = self._expected_presigned_url
+
+    def call_create_presigned_get(self) -> None:
+        """Calls the `ImageStore` `create_presigned_get` method with the appropriate data from a prior call to
+        `mock_create_presigned_get`."""
```
rowan04, line 110:
```suggestion
        """
        Calls the `ImageStore` `create_presigned_get` method with the appropriate data from a prior call to
        `mock_create_presigned_get`.
        """
```

test/unit/stores/test_image.py
```diff
@@ -82,3 +83,56 @@ def test_upload(self):
...
+    _image_out: ImageOut
+    _expected_presigned_url: str
+    _obtained_presigned_url: str
+
+    def mock_create_presigned_get(self, image_in_data: dict) -> None:
+        """
+        Mocks object store methods appropriately to test the `create_presigned_get` store method.
+
+        :param image_in_data: Dictionary containing the image  data as would be required for an
+        `ImageIn`.
```
rowan04, line 100:
```suggestion
        :param image_in_data: Dictionary containing the image data as would be required for an
            `ImageIn`.
```

rowan04 (commented):
Some more minor docstring changes

VKTB (dismissed):

object_storage_api/core/exceptions.py
asuresh-code: For the exception and schemas docstring formatting, I was copying the existing templates in the `object_storage_api`. @joelvdavies I can change the rest of the exception and schemas to 1 line docstrings as well, unless there was a reason they were set like this?

joelvdavies: No reason just copied from IMS, so yeah you can change them.

test/unit/stores/test_image.py
```diff
@@ -106,8 +106,9 @@ def mock_create_presigned_get(self, image_in_data: dict) -> None:
         self.mock_s3_client.generate_presigned_url.return_value = self._expected_presigned_url
 
     def call_create_presigned_get(self) -> None:
-        """Calls the `ImageStore` `create_presigned_get` method with the appropriate data from a prior call to
-        `mock_create_presigned_get`."""
+        """
+        Calls the `ImageStore` `create_presigned_get` method with the appropriate data from a prior call to
+            `mock_create_presigned_get`."""
```
joelvdavies, line 111:
```suggestion
            `mock_create_presigned_get`.
"""
```

joelvdavies (approved):

rowan04 (approved):

PR #70, opened 2024-11-25
Add production Dockerfile and ci image upload workflow

### LINKED ISSUE #5: Add production Dockerfile and upload images on CI
- [ ] Add `Dockerfile.prod`
- [ ] Add step to `.ci.yml` that builds and uploads the prod image on PRs like `inventory-management-system-api` does

asuresh-code:
## Description
- Updates the `Dockerfile` to have multistages:
  - `dev` for local development
  - `test` for running the tests
  - `prod` for production
- Updates the test CI jobs to run the tests in a Docker container
- Updates the `docker` CI job to use a workflow dispatch input so that it does not push an image to Harbor on every push to a PR but only when the [CI workflow is manually run](https://github.com/ral-facilities/object-storage-api/actions/workflows/.ci.yml) 

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #5


Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
+FROM python:3.12.7-alpine3.20@sha256:edd1d8559c585e1e9a9b79de44ac27f8ac32cb0c7323e112ae6870ceeecd8dbf AS builder
```
VKTB, line 1: You can probably get away with one stage as I don't think it would make a massive difference in this case.

Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
...
+    python3 -m pip install --no-cache-dir -r requirements.txt; 
+
+FROM python:3.12.7-alpine3.20@sha256:edd1d8559c585e1e9a9b79de44ac27f8ac32cb0c7323e112ae6870ceeecd8dbf
+
+WORKDIR /object-storage-api-run
+
+COPY --from=builder /usr/local/lib/python3.12/site-packages /usr/local/lib/python3.12/site-packages
+COPY --from=builder /usr/local/bin /usr/local/bin
+
+COPY README.md ./
```
VKTB, line 17: This is not needed.

Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
...
+COPY object_storage_api/ object_storage_api/
+
+RUN set -eux; \
+    \
+    # Create loging.ini from its .example file \
+    cp object_storage_api/logging.example.ini object_storage_api/logging.ini;
```
VKTB, line 23: We don't do this in other projects. We have instructions to tell the user to create the file manually before building the image. I think we should stay consistent.

Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
...
+
+USER nobody
```
VKTB, line 25: Wasn't aware of this user. We just create a non-root user and group in other projects. I will need to look into this a bit more because if we can just use `nobody` then it would save us from creating non-root users. Let me know if you did any research on this and what you find out.

Dockerfile.prod
```diff
@@ -0,0 +1,28 @@
...
+
+CMD ["uvicorn", "object_storage_api.main:app", "--app-dir", "/object-storage-api-run", "--host", "0.0.0.0", "--port", "8000"]
```
VKTB, line 27: We recently started using the [FastAPI CLI](https://fastapi.tiangolo.com/fastapi-cli/) CLI command line program, which internally uses a production-ready Uvicorn server, as that now seems to be the recommended way by FastAPI of running apps for, both, local development and production.

Dockerfile.prod
VKTB, line 1: You might want to look at the `Dockerfile` in https://github.com/ral-facilities/scigateway-auth/pull/134 and what I did there. If you like what I did then it would be nice to keep things consistent (thoughts on that PR are welcome).

VKTB (commented):
Still waiting on the Harbor project. The README will need to be updated at the end to include instructions.

.github/workflows/.ci.yml
```diff
@@ -106,3 +106,34 @@ jobs:
       - name: Output docker logs (minio)
         if: failure()
         run: docker logs object-storage-api-minio-1
+  docker:
```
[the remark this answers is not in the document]
asuresh-code: I've made those changes to this PR, although I can't see if the image is getting pushed to Harbor in the logs? It may be because this is a draft PR; I have tried pushing a tag to see if that triggers it.

Dockerfile.prod
asuresh-code: Ah, I saw we did this in the IMS api repo, so I assumed we were following that model.

.github/workflows/.ci.yml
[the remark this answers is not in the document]
VKTB: If you have a look at the description of https://github.com/ral-facilities/scigateway-auth/pull/134, you will note that it does not push an image on every push to a PR but will only do when this workflow is manually run from the [Actions tab](https://github.com/ral-facilities/object-storage-api/a
ci
ctions/workflows/.ci.yml). No tag is required for this.

Dockerfile.prod
VKTB: My bad, we do this in other projects, you are right. However, I think this is unnecessary because the instructions tell the user to create the file manually before building the image. The [new SciGateway Auth Dockerfile](https://github.com/ral-facilities/scigateway-auth/blob/port-over-to-fastapi-%23133/Dockerfile) does not copy this in and at some point, I am going to refactor IMS API and LDAP JWT AUth to be consistent with SciGateway Auth and have only one Dockerfile that do not copy this in and have stages for dev and prod. I suggested in my other comment that you could do the same for this repo.

Dockerfile.prod
asuresh-code: I had a look at the PR, and I like the idea of keeping of merging the different dockerfiles into 1 file with multiple stages. My only suggestion, is if it was implemented here for us to make a script to run everything succintly.

Looking at the README of that PR, the commands to build the image, and run tests/start the container are very long, so it would be useful to have shortened commands like we do on IMS.

VKTB: > My only suggestion, is if it was implemented here for us to make a script to run everything succintly.

Looking at the README of that PR, the commands to build the image, and run tests/start the container are very long, so it would be useful to have shortened commands like we do on IMS.

Sorry, I am not sure I fully understand what you mean. Could you please explain with examples if possible?

asuresh-code: image
It's a small point, that the command for running the containers seem like they are very long (compared to just using docker compose up), and I'm not sure if you need to build an image each time you switch from testing to developing or make changes? I think you could use a `docker-compose.yml` to run them (shortening the commands), although I'm not sure if/how you would configure it to work with different targets for the same dockerfile (i.e docker compose test up, docker compose dev up, etc)

VKTB: Thanks for your explanation, I see what you mean now. This is why in the `README`, the _Using `docker-compose.yml` for local development_ section is at the top of the _Inside of Docker_ section, telling the reader that this is the easiest way to run the app for local development.

As for testing, given that this app requires other services like MongoDB, minio etc to be spun up for the e2e to run, it makes sense to have a second file called `docker-compose.test.yml` file. In that file you can define the services that are needed for testing and set the `object-storage-api` container to use the `test` target from the `Dockerfile`. You can then use the `docker compose up` and `docker compose down` to run the tests using that new file. At the same time, you should rename the current `docker-compose.yml` file to `docker-compose.dev.yml` and only use that for local development. As both Docker Compose files have the source code mounted through volumes, you will not need to rebuild an image when you make changes to the code.

SciGateway Auth doesn't require any services to be spun up so it is easier (at least for me) to just run the tests with the `docker run` command as I can just copy and paste it.

asuresh-code: Thanks for the response, it helped clarify a lot of points. I've made all of the changes covered here now, but I had a few clarifying questions to ask:
- Does prod also need its own compose file? I assume it doesn't, and that you only need the stage to push to Harbor.
- In the dockerfile test stage, I'm not sure if the `CMD` line should run `fastapi` or `pytest`. It's currently running pytest (as I can see you did the same in the FastAPI PR), but I don't really understand why this decision is made? If I use the `docker-compose.test.yml` file the tests don't automatically run?

asuresh-code: In the latest commit I have an example implementation of using Environment Variables for the compose file.

I do agree if there would be more differences in the future, then keeping them separate sounds like a good idea since both commands are of similar length, and there wouldn't likely be more compose files in the future.

VKTB: > Regarding the tests not running, I will have a look next week because they should do if everything is configured correctly.

I spent too long on this today.. I discovered that Docker Compose doesn't respect the target if there is already a Docker image for the container. For example, if I run `docker compose -f docker-compose.dev.yml up`, it will build an image form the `object-storage-api` container and spin up a container that will start the app. If I then stop the containers (using `Ctrl_c`) and run `TARGET_STAGE=test docker compose -f docker-compose.dev.yml up`, it will not build a new image for the the `object-storage-api` but use the one it built before which is why it doesn't run the tests. Rebuilding the images (`docker compose -f docker-compose.dev.yml up`) before running the Docker Compose `up` command each time the target is switched seems to solve this problem.

However, this issue does not arise (and no rebuilding of images is required) if the two compose files are used and different service and container names are used in each of the files. For example, naming the service `object-storage-api-test` and the container `object_storage_api_test_container` in the `docker-compose.test.yml`.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.12.8-alpine3.20@sha256:0c4f778362f30cc50ff734a3e9e7f3b2ae876d8386f470e0c3ee1ab299cec21b
+FROM python:3.12.8-alpine3.20@sha256:0c4f778362f30cc50ff734a3e9e7f3b2ae876d8386f470e0c3ee1ab299cec21b as base
```
VKTB, line 1: You don't really need a base stage in this case. You can just move this to the `dev` stage considering that the `tests` stage uses it as the base.

Dockerfile
```diff
@@ -10,5 +10,41 @@ RUN --mount=type=cache,target=/root/.cache \
...
 CMD ["fastapi", "dev", "object_storage_api/main.py", "--host", "0.0.0.0", "--port", "8000"]
 EXPOSE 8000
+
+FROM dev as test
+
+WORKDIR /object-storage-api-run
+
+COPY test/ test/
+
+CMD ["pytest", "--config-file", "test/pytest.ini", "test/", "--cov"]
```
VKTB, line 30: This would fail because only the prod dependencies are installed.

.github/workflows/.ci.yml
```diff
@@ -90,11 +95,11 @@ jobs:
...
-          docker compose up -d mongo-db minio minio_create_buckets
+          TARGET_STAGE=test docker compose -f docker-compose.dev.yml up -d mongo-db minio minio_create_buckets
           sleep 10
       - name: Create MinIO buckets
         run: |
-          docker compose up minio_create_buckets
+          TARGET_STAGE=test docker compose -f docker-compose.dev.yml up minio_create_buckets
 
       - name: Run e2e tests
         run: pytest -c test/pytest.ini test/e2e/ --cov
```
VKTB, line 105: Tests should now run in a Docker container. This requires some thinking given there are multiple jobs for the tests.

VKTB (changes requested):

Dockerfile.prod
asuresh-code: Ah I see, well in that case it seems even more appropriate to keep separate compose files in this case.

.github/workflows/.ci.yml
asuresh-code: 1 solution I've thought of is in the `Dockerfile`, we can split up the build stage for test into a unit test stage, and e2e test stage (the only difference being the command run). 
image

Then in the workflow in the `run e2e/unit test` step of the separate jobs, we build the image targeting the unit or e2e stage respectively, and then run the image instead of using the docker compose file.

```
docker build --target unit-test -t object-storage-api:unit-test .
docker run object-storage-api:unit-test
```

This saves us having to create a new compose file for unit/e2e testing, or introduce environment variables into the compose file. This does create a new problem of now the docker-compose.test.yml file no longer has a target stage that runs all of the tests. This could be addressed by:

1. Define a 3rd test stage in the Dockerfile that runs all tests (same as existing test stage essentially), and set that as the target for the compose file
2. Create 2 separate services + containers in the compose file 1 for unit testing and 1 for e2e testing. The separate jobs can now just run each service to their respective tests, and to run all tests you can just run everything? I'm less sure about this approach, if the services would conflict with each other if you try to run them at the same time.

```
services:
  object-storage-api-unit-test:
    container_name: object_storage_api_container_unit_test
    build: 
      context: .
      target: unit-test
    ...

  object-storage-api-e2e-test:
    container_name: object_storage_api_container_e2e_test
    build: 
      context: .
      target: e2e-test
    ...
```
`docker-compose -f docker-compose.test.yml up object-storage-api-unit-tests`

VKTB:
@asuresh-code Thanks for all your work on this PR so far, as agreed I will continue working on it.

VKTB:
This PR does require a bit of work before it can be merged but I need a production Docker image so I created a new PR (https://github.com/ral-facilities/object-storage-api/pull/107) that addresses just this.

VKTB:
The `docker compose up` command does not exit if other containers continue running so having a `docker-compose.test.yml` and using it in the CI to run the tests means the step will never complete. Locally, you will have to force it to exit so I did not see the point of having a `docker-compose.test.yml`. Instead, the tests will need to be run as per instructions on the README.

VKTB:
@asuresh-code This is now ready to be reviewed.

asuresh-code (commented):
Looks good, tried using the commands in the README and everything worked well. Just had a question about one of the docker run commands.

.github/workflows/.ci.yml
```diff
@@ -77,44 +73,38 @@ jobs:
...
+          docker compose up minio-create-buckets
 
       - name: Run e2e tests
-        run: pytest -c test/pytest.ini test/e2e/ --cov
+        run: |
+          docker build --file Dockerfile --target test --tag object-storage-api:test .
+          docker run \
+           --name object-storage-api \
+           --add-host localhost:host-gateway \
+           object-storage-api:test \
```
asuresh-code, line 94: I wasn't sure what the `--add-host localhost:host-gateway` line is doing, and why it is needed for the e2e tests, but not the unit tests?

asuresh-code: Ah, thanks for pointing me to the comment, I somehow missed that the first time :sweat_smile:

Yep in that case it LGTM, happy for this to be approved and merged.

VKTB: No worries at all. Thank you for your review!

PR #72, opened 2024-11-28
Add delete endpoint for image

### LINKED ISSUE #38: Add an endpoint for deleting an image
Implement a DELETE endpoint `/images/<image_id>` that:

- [x] Deletes the image metadata from the database
- [x] Deletes the file object in the object store
- [x] Must gracefully deal with a situation when the file object with the given `object_key` doesn't exist

asuresh-code:
## Description
Implements a DELETE endpoint `/images/<image_id>` that:

- Deletes the image metadata from the database
- Deletes the file object in the object store
- Gracefully deals with a situation when the file object with the given `object_key` doesn't exist

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #38


joelvdavies (commented):
A few small comments. will look at the tests afterwards. This PR may have some minor conflicts with the get branch so will probably need merging into to this branch one it is merged.

object_storage_api/stores/image.py
```diff
@@ -37,3 +37,16 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
...
+
+    def delete(self, object_key: str) -> None:
+        """
+        Deletes a given image from object storage.
+
+        :param object_key: Key of the image to delete.
+        """
+
+        logger.info("Deleting image file from the object storage")
+        s3_client.delete_object(
```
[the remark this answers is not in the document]
joelvdavies: Good to know thanks.

object_storage_api/services/image.py
```diff
@@ -85,3 +85,13 @@ def list(self, entity_id: Optional[str] = None, primary: Optional[bool] = None)
...
+
+    def delete(self, image_id: str) -> None:
+        """
+        Delete an image by its ID.
+
+        :param image_id: The ID of the image to delete.
+        """
+
+        object_key = self._image_repository.delete(image_id)
+        self._image_store.delete(object_key)
```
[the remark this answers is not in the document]
joelvdavies: Its just because these tasks have ` Must gracefully deal with a situation when the file object with the given object_key doesn't exist` in them so if its deleted first and it ends up not having the file then that is considered ok. But we dont want the other way round as the way we would look up things in the object storage is using the database.

Also the transaction wouldn't work the other way round as its a MongoDB thing - in this case it would ensures that if an error occurs before it is finished, none of the changes actual end up committed to the database, so in this case if it was wrapped in a transaction if self._image_store.delete raised any error the first database edit would never be finalised.

test/unit/stores/test_image.py
```diff
@@ -85,6 +85,33 @@ def test_upload(self):
...
+        self.mock_s3_client.delete_object.assert_called_once_with(
+            Bucket=object_storage_config.bucket_name.get_secret_value(),
+            Key="object_key",
+        )
+
+
+class TestDelete(DeleteDSL):
+    """Tests for deleting an image from object storage."""
+
+    def test_create_presigned_post(self):
```
joelvdavies, line 108:
```suggestion
    def test_delete(self):
```

test/unit/stores/test_image.py
```diff
@@ -85,6 +85,33 @@ def test_upload(self):
         self.check_upload_success()
 
 
+class DeleteDSL(ImageStoreDSL):
+    """Base class for `delete` tests."""
+
+    def call_delete(self) -> None:
+        """Calls the `ImageStore` `delete` method."""
+
+        self.image_store.delete("object_key")
```
joelvdavies, line 94: Not necessary but for consistency I would prefer taking in the image_id, store it in a `_delete_system_id: str` and then assert using that like we do for repo tests to be consistent.

test/unit/services/test_image.py
```diff
@@ -247,3 +247,33 @@ def test_list(self):
...
+        :param image_id: ID of the image to be deleted.
+        """
+
+        self._delete_image_id = image_id
+        self.image_service.delete(image_id)
+
+    def check_delete_success(self) -> None:
+        """Checks that a prior call to `call_delete` worked as expected."""
+
+        self.mock_image_repository.delete.assert_called_once_with(self._delete_image_id)
```
joelvdavies, line 270: Shouldn't this also assert the store delete was called?

joelvdavies (commented):
Minor comments works well.

test/unit/services/test_image.py
asuresh-code: It should be, thank you for pointing it out.

test/unit/services/test_image.py
```diff
@@ -247,3 +247,45 @@ def test_list(self):
...
+    def call_delete(self) -> None:
+        """Calls the `ImageService` `delete` method."""
+
+        self.image_service.delete(self._delete_image_id)
+
+    def check_delete_success(self) -> None:
+        """Checks that a prior call to `call_delete` worked as expected."""
+
+        self.mock_image_repository.delete.assert_called_once_with(self._delete_image_id)
+        self.mock_image_store.delete.assert_called_once_with(self._delete_image_object_key)
```
joelvdavies, line 281:
```suggestion
        self.mock_image_store.delete.assert_called_once_with(self._delete_image_object_key)
        self.mock_image_repository.delete.assert_called_once_with(self._delete_image_id)
```
Just to match the call order.

joelvdavies (dismissed):
LGTM

test/e2e/test_image.py
```diff
@@ -307,3 +307,63 @@ def test_list_with_invalid_primary_filter(self):
...
+
+        assert self._delete_response_image.status_code == 204
+
+    def check_delete_image_failed_with_detail(self) -> None:
+        """
+        Checks that a prior call to `delete_image` gave a failed response with the expected code and
+        error message.
+
+        :param status_code: Expected status code of the response.
+        :param detail: Expected detail given in the response.
```
VKTB, line 338:
```suggestion
```
if the method is not meant to take these in.

test/e2e/test_image.py
```diff
@@ -307,3 +307,63 @@ def test_list_with_invalid_primary_filter(self):
...
+        Deletes an image with the given ID.
+
+        :param image_id: ID of the image to be deleted.
+        """
+
+        self._delete_response_image = self.test_client.delete(f"/images/{image_id}")
+
+    def check_delete_image_success(self) -> None:
+        """Checks that a prior call to `delete_image` gave a successful response with the expected data
+        returned."""
```
VKTB, line 328:
```suggestion
        """Checks that a prior call to `delete_image` gave a successful response with the expected data returned."""
```

object_storage_api/routers/image.py
```diff
@@ -78,3 +78,18 @@ def get_image(
     logger.info("Getting image with ID: %s", image_id)
 
     return image_service.get(image_id)
+
+
+@router.delete(
+    path="/{image_id}",
+    summary="Delete an image by ID",
+    response_description="Image deleted sucessfully",
```
VKTB, line 86:
```suggestion
    response_description="Image deleted successfully",
```

VKTB (approved):

PR #73, opened 2024-11-28
Add get endpoint for a list of attachment metadata #25

### LINKED ISSUE #25: Add an endpoint for getting a list of attachment metadata
Implement a GET endpoint `/attachments` that:
- [ ] Returns a list of attachments
  - Should include the metadata without `object_key` 
- [ ] Can be given an optional query parameter `entity_id` that filters results by the `entity_id` via the database

rowan04:
## Description

Has optional query parameter `entity_id` that filters results

TO DO: add tests

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Resolves #25


rowan04:
As referenced in my last commit, the get request would fail because `AttachmentPostResponseSchema` was missing the `upload_info` argument. Did I fix this the correct way?

rowan04:
tests to be added

joelvdavies:
> As referenced in my last commit, the get request would fail because `AttachmentPostResponseSchema` was missing the `upload_info` argument. Did I fix this the correct way?

It needs a new schema for the get request as the one you mentioned is for the post response which includes a url to upload the attachment to. Similar to https://github.com/ral-facilities/object-storage-api/pull/69 I would recommend calling it `AttachmentMetadataSchema` and inherit from `AttachmentPostSchema`, add the `id` and then inherit `AttachmentMetadataSchema` in `AttachmentPostResponseSchema` to add the `upload_info`.

joelvdavies:
> like this @joelvdavies?

Yes exactly.

rowan04:
What to do about the failing Linting check?

joelvdavies:
> What to do about the failing Linting check?

I would put
```python
# Expect some duplicate code inside tests as the tests for the different entities can be very similar
# pylint: disable=duplicate-code
```
at the top of the image and attachment repo unit test files under the first comment like we have done in `test_catalogue_category.py` on ims-api.

asuresh-code (changes requested):
Looks Good overall, and works well, just have a few comments

test/unit/repositories/test_attachment.py
```diff
@@ -86,3 +91,68 @@ def test_create(self):
...
+    _entity_id_filter: Optional[str]
+    _obtained_attachment_out: list[AttachmentOut]
+
+    def mock_list(self, attachment_in_data: list[dict]) -> None:
+        """
+        Mocks database methods appropriately to test the `list` repo method.
+
+        :param attachment_in_data: List of dictionaries containing the attachment data as would be required for an
+            `AttachmentIn` database model (i.e. no ID or created and modified times required).
+        """
```
asuresh-code, line 109:
```suggestion
            `AttachmentIn` database model (i.e. no created and modified times required).
        """
```
`AttachmentIn` models actually do contain the ID already. There's a similar error in the `test/unit/repositories/test_image.py` file if you could fix that as well

asuresh-code:
Currently an invalid `entity_id` returns a `422` status_code and an error message saying it received an `Invalid ID`. In other endpoints we have moved to returning a `404` and an error message saying `BLANK was not found` when an ID is invalid. I believe it's for security reasons, and not letting outsiders figure out our ID formatting?

Also, for more recent endpoints like the [get singular for images ](https://github.com/ral-facilities/object-storage-api/pull/69), we've moved the error handling to the repo layer, so I think you should implement it there, unless @joelvdavies has other ideas?

test/e2e/test_attachment.py
```diff
@@ -132,3 +134,123 @@ def test_create_with_invalid_entity_id(self):
...
+        """
+        Checks that a prior call to `get_attachments` gave a successful response with the expected data returned.
+
+        :param expected_attachments_get_data: List of dictionaries containing the expected attachment data as would
+            be required for an `AttachmentMetadataSchema`.
+        """
+        assert self._get_response_attachment.status_code == 200
+        assert self._get_response_attachment.json() == expected_attachments_get_data
+
+    def check_attachments_list_response_failed_with_message(self, status_code, expected_detail, obtained_detail):
```
joelvdavies, line 202:
```suggestion
    def check_get_attachments_failed_with_message(self, status_code, expected_detail, obtained_detail):
```
Please also modify the one for images `check_images_list_response_failed_with_message` to match - they should be the same name as the success/request function itself.

joelvdavies (commented):
Just a few other minor things.

joelvdavies:
> > Currently an invalid `entity_id` returns a `422` status_code and an error message saying it received an `Invalid ID`. In other endpoints we have moved to returning a `404` and an error message saying `BLANK was not found` when an ID is invalid. I believe it's for security reasons, and not letting outsiders figure out our ID formatting?
> 
> Yes in this case if you get a route with a non existent ID it makes sense for it be a 404 anyway as the code signifies that the route and entity wasn't found, normally a 422 would be used if something in the request body schema itself is invalid. Like Anikesh said I believe it was also because you could figure out which ids were valid and invalid by looking at the error messages.

Sorry, I have thought about it a bit more and realised I made a mistake here 🤦‍♂️ . Given these are query parameters and not path parameters it does make sense to be that they are 422s and not 404s. So the current behaviour is fine, the only thing we could do in reference to this is return an empty `[]` list like ims-api does in such cases. I would just create another issue to potentially do that afterwards, no need to do it here.

rowan04:
> Also, for more recent endpoints like the [get singular for images ](https://github.com/ral-facilities/object-storage-api/pull/69), we've moved the error handling to the repo layer, so I think you should implement it there, unless @joelvdavies has other ideas?

Hi @joelvdavies and @asuresh-code, I've implemented what I think you mean with this and error status codes, but i'm not sure! If what I've done is correct, it needs doing for images too, as there is no error handling in the repositories layer for list images.

joelvdavies:
> > Also, for more recent endpoints like the [get singular for images ](https://github.com/ral-facilities/object-storage-api/pull/69), we've moved the error handling to the repo layer, so I think you should implement it there, unless @joelvdavies has other ideas?
> 
> Hi @joelvdavies and @asuresh-code, I've implemented what I think you mean with this and the error status codes, but i'm not sure! If what I've done is correct, it needs doing for images too, as there is no error handling in the repositories layer for list images.

Sorry as I mentioned above I think what you had before for the error was fine, so I would just go back to that for now, then we can have another separate issue to change it to be like ims-api, the only difference there that there we return `[]`, while the list response for images and here returned a 422 saying its an invalid id. So I would just revert the last commit above.

asuresh-code (approved):
LGTM 🎉 good work!

VKTB (approved):
LGTM 🎉

VKTB:
> then we can have another separate issue to change it to be like ims-api, the only difference there that there we return [], while the list response for images and here returned a 422 saying its an invalid id.

Please open an issue for this if one has not already been opened when this PR gets merged, thanks.

Issue #81, opened 2024-12-19
Single + Multiple Download for Images


asuresh-code:
Mirroring this [front-end issue](https://github.com/ral-facilities/inventory-management-system/issues/1078), the back-end needs to allow for images to be downloaded instead of displayed just inline.

Also in line with this feature, file names need to be parsed against their file type to ensure the correct file extension is in place; otherwise there are issues trying to open downloaded images. There are 3 main solutions:

1. **Store Image MIME type in separate column**
2. **Store Image file extension in separate column**
3.  **Store image file extension with filename**

In this PR I've opted for the 1st option. Options 1 and 2 (Storing an extra column) would also allow us to show the image type in the image info, and allow filtering by type. They also seem more robust than Option 3, which would involve string manipulation to remove and add the file extension back in after image edits. 

Between Option 1 and 2, option 2 would be better if we did not want additional enhancements (like viewing and filtering by image type); With Option 2, you would extract the MIME type from the header, convert it to an extension once and store it, where it would be appended to the download url's filename. Whereas Option 1, you would store the MIME type and convert it to an extension everytime an image was downloaded. However, if the filtering/viewing enhancements were implemented it would make more sense to use Option 1.

All of the options are viable however, so any of the implementations work fine.

joelvdavies:
I agree option 1 is more sensible out of these. Just not quite sure why this is needed? The content type is already stored in the object store so I was under the impression all that was needed was an additional parameter in the presigned url to make it download. The content type actually comes from the extension anyway, as the front end determines that when it uploads.

asuresh-code:
> The content type is already stored in the object store so I was under the impression the most that was needed was an additional parameter in the presigned url to make it download

Although the content-type is set in the object store, it doesn't actually seem to affect the file type of the image downloaded (at least on a Windows System). The file extension seems to override what the perceived file type is, so if you download a file called `image.mp4` and set the content type to `"image/jpeg"` both on upload and download, it will still be perceived as an MP4 file. I don't think it's an issue with our implementation, it may be a quirk of generating a presigned url instead of just getting the object.

> download but I even thought that may be possible on the front end as the data is returned from the view anyway

I think you're right for this case; I believe you can set the `ResponseContentDisposition:` as `inline` for both viewing and downloading images, but I didn't investigate that yet. However, setting it to `attachment` for both viewing and downloading images, seems to let you do both as well.

> The content type actually comes from the extension anyway, as the front end determines that when it uploads.

[...]

PR #82, opened 2024-12-19
Single/Multiple Image Download

asuresh-code:
## Description

See #81 Implements Option 1 (Store Mime Type)

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to #81


object_storage_api/stores/image.py
joelvdavies, line 1: Just a query, is your formatter set to line-length 120, there are two lines here that were unchanged and below 120, but still got formatted?

object_storage_api/stores/image.py
```diff
@@ -46,13 +51,17 @@ def create_presigned_get(self, image: ImageOut) -> str:
...
+            "Generating presigned url to get image with object key: %s from the object store",
+            image.object_key,
+        )
         response = s3_client.generate_presigned_url(
             "get_object",
             Params={
                 "Bucket": object_storage_config.bucket_name.get_secret_value(),
                 "Key": image.object_key,
                 "ResponseContentDisposition": f'inline; filename="{image.file_name}"',
+                "ResponseContentType": "application/octet-stream",
```
joelvdavies, line 64: Does this have to be done here? I think looking at https://stackoverflow.com/questions/20508788/do-i-need-content-type-application-octet-stream-for-file-download we want the last case, keeping the original format by leaving it to the one in the object-store here.

object_storage_api/services/image.py
```diff
@@ -58,6 +59,12 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
         # Upload the full size image to object storage
         object_key = self._image_store.upload(image_id, image_metadata, upload_file)
 
+        expected_file_type = mimetypes.guess_type(upload_file.filename)[0]
+        if expected_file_type != upload_file.content_type:
+            raise InvalidObjectIdError(
```
joelvdavies, line 64: I think this needs to be a new type of exception to indicate the content type. It would still be a 422 though.

test/unit/services/test_image.py
```diff
@@ -151,14 +153,24 @@ class TestCreate(CreateDSL):
     def test_create(self):
         """Test creating an image."""
 
-        self.mock_create(IMAGE_POST_METADATA_DATA_ALL_VALUES)
+        self.mock_create(IMAGE_POST_METADATA_DATA_ALL_VALUES, "test.png")
         self.call_create()
         self.check_create_success()
 
+    def test_create_with_invalid_file(self):
```
joelvdavies, line 160:
```suggestion
    def test_create_with_file_extension_content_type_mismatch(self):
```
I think it could do with an e2e test too if possible.

object_storage_api/services/image.py
asuresh-code: Agreed, I did actually make a new exception, but forgot to push 😵‍💫

object_storage_api/stores/image.py
asuresh-code: I think there's probably a way to do without it. I was under the impression that application/octet-stream doesn't override it, but seems like that isn't the case.

object_storage_api/stores/image.py
asuresh-code: It was set in the workplace, and it I just set it in the user settings and still reformats the line. For some reason, if I delete the trailing comma in both of the lines they don't get reformatted, and stay on one line?

joelvdavies: Ah yeah it does that if you intentionally put a comma at the end it wont auto remove it and just leaves it on multiple, whereas without it it reformats.

test/unit/services/test_image.py
```diff
@@ -160,14 +162,24 @@ class TestCreate(CreateDSL):
...
```
joelvdavies, line 169: Could update this test name so its more explicit about what is invalid (e.g. example above).

test/e2e/test_image.py
```diff
@@ -364,6 +364,12 @@ def test_update_invalid_id(self):
         self.patch_image("invalid-id", {})
         self.check_patch_image_failed_with_detail(404, "Image not found")
 
+    def test_partial_update_with_mismatched_extension(self):
+        """Test updating an image with a different extension."""
+        image_id = self.post_image(IMAGE_POST_METADATA_DATA_ALL_VALUES, "image.jpg")
+        self.patch_image(image_id, {**IMAGE_PATCH_METADATA_DATA_ALL_VALUES, "file_name": "picture.png"})
+        self.check_patch_image_failed_with_detail(422, "Filename does not contain the correct extension")
```
joelvdavies, line 371:
```suggestion
        self.check_patch_image_failed_with_detail(422, "Filename does not contain the correct extension for the content type")
```

joelvdavies:
One other thing could you perhaps update the names of this PR and the issue to reflect its new purpose of preventing invalid files/add the issue to the IMS project board please, just helps us find things again.

rowan04:
This is also an issue on the attachments side - I have created an issue for it #96

rowan04 (approved):
lgtm!

joelvdavies (dismissed):

object_storage_api/services/image.py
```diff
@@ -63,6 +65,12 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
         # Upload the full size image to object storage
         object_key = self._image_store.upload(image_id, image_metadata, upload_file)
 
+        expected_file_type = mimetypes.guess_type(upload_file.filename)[0]
+        if expected_file_type != upload_file.content_type:
+            raise InvalidFilenameExtension(
+                f"File extension `{upload_file.filename}` does not match content type `{upload_file.content_type}`"
+            )
```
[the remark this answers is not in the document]
joelvdavies: Yep good spot, it shouldn't upload unless its valid.

PR #86, opened 2025-01-03
Force auth before UploadFile upload #85

### LINKED ISSUE #85: Authentication happens after upload for images
See https://github.com/fastapi/fastapi/issues/4941. I think we need to use a middleware instead of the router dependencies as it appears the `UploadFile` is fully formed prior to the router dependency being executed.

joelvdavies:
## Description

This solution is not ideal. Dependencies tend to be preferred where possible and in this case future work to add specific user permissions likely would still need a dependency to decode the token leading to it being decoded twice. Also in order to get the docs to work, their paths have to be specifically excluded from authentication while also ensuring a `HTTPBearer` dependency is added to the routers for the `Authorize` button to show up even though it instead uses the middleware.

The unit tests have been removed since the middleware requires an the FastAPI app to be running to function, so have replaced with the e2e tests.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #85


object_storage_api/auth/jwt_middleware.py
```diff
@@ -17,37 +21,39 @@
...
-        credentials: HTTPAuthorizationCredentials = await super().__call__(request)
-
-        if not self._is_jwt_access_token_valid(credentials.credentials):
-            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Invalid token or expired token")
-
-        return credentials.credentials
+        if (
+            not request.url.path == f"{config.api.root_path}/docs"
+            and not request.url.path == f"{config.api.root_path}/openapi.json"
+        ):
```
VKTB, line 44: You can save a few lines by doing
```suggestion
        if request.url.path not in [f"{config.api.root_path}/docs", f"{config.api.root_path}/openapi.json"]:
```

VKTB (approved):

VKTB (approved):

VKTB:
Just noticed that the docs are not actually excluded from the authentication when `API__ROOT_PATH` is set. Will be addressed as part of https://github.com/ral-facilities/object-storage-api/issues/114.

PR #88, opened 2025-01-07
Edit Image Primary Field

### LINKED ISSUE #52: Allow editing of an image's `primary` field
Following on from https://github.com/ral-facilities/object-storage-api/issues/35. Also

- [x] Allow editing of `primary`
- [x] Ensure that when an image is made primary - all others have `primary` set to False - it seems it will be possible without transactions if using a `bulkWrite` (preferable) or aggregation (while selecting just those with `primary: True`, or the correct ID to reduce the match stage memory size)

asuresh-code:
## Description
See #52 

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #52

### DIFF
--- a/object_storage_api/repositories/image.py
+++ b/object_storage_api/repositories/image.py
@@ -31,7 +31,7 @@ class ImageRepo:
         self._database = database
         self._images_collection: Collection = self._database.images
 
-    def create(self, image: ImageIn, session: ClientSession = None) -> ImageOut:
+    def create(self, image: ImageIn, session: Optional[ClientSession] = None) -> ImageOut:
         """
         Create a new image in a MongoDB database.
 
@@ -44,7 +44,7 @@ class ImageRepo:
         result = self._images_collection.insert_one(image.model_dump(by_alias=True), session=session)
         return self.get(str(result.inserted_id), session=session)
 
-    def get(self, image_id: str, session: ClientSession = None) -> ImageOut:
+    def get(self, image_id: str, session: Optional[ClientSession] = None) -> ImageOut:
         """
         Retrieve an image by its ID from a MongoDB database.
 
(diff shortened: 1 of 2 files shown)


asuresh-code:
I noticed that currently you cannot set the primary value when you upload/post an image. Is that meant to be like that in production, or missed/waiting for this issue to be done first?

If it should be added, I can add the back-end changes into this PR as well.

joelvdavies:
> I noticed that currently you cannot set the primary value when you upload/post an image. Is that meant to be like that in production, or missed/waiting for this issue to be done first?

Yes this was the intention, depending on how the front is done I think we expected it would end up separate process after an upload to pick which one should be primary, or to update it. If this changes it can be added separately.

> If it should be added, I can add the back-end changes into this PR as well.
> 
> Also, the missing line in the codecov report is covered by the E2E test.

Can it not be covered though a unit test as well? Its hard to see coverage without it there.

asuresh-code:
> Can it not be covered though a unit test as well? Its hard to see coverage without it there.

I can get the line coverage easily in a unit test, although I'm not sure if I would be able to test if the query actually set other images primary value to False? Since the mongodb queries are being mocked, I initially thought there wasn't any point in covering these lines, since I thought there entire functionality was being mocked, although I can see it's a good idea to include them anyway since you can check to make sure the bulkwrite is correctly formed.

joelvdavies (changes requested):
Worked well, just two comments.

joelvdavies (approved):
Works well.

object_storage_api/repositories/image.py
```diff
@@ -110,9 +112,18 @@ def update(self, image_id: str, image: ImageIn, session: ClientSession = None) -
         logger.info("Updating image metadata with ID: %s", image_id)
         try:
             image_id = CustomObjectId(image_id)
-            self._images_collection.update_one(
-                {"_id": image_id}, {"$set": image.model_dump(by_alias=True)}, session=session
-            )
+            if update_primary:
+                bulkwrite_update = [
+                    UpdateMany(
+                        filter={"primary": True, "entity_id": image.entity_id}, update={"$set": {"primary": False}}
```
VKTB, line 118: How many images can have a value of `True` for the primary field at any one time? My understanding was just one? If that is the case then should we be using `UpdateMany` here if only one needs to be updated or are we going with this for a peace of mind?

VKTB: That's fine then, thanks for clarifying.

VKTB (dismissed):

PR #91, opened 2025-01-09
Add get endpoint for attachment metadata with download url #26

### LINKED ISSUE #26: Add an endpoint for getting an attachment with a download URL
Implement a GET endpoint `/attachments/<attachment_id>` that:
- [ ] Generates a pre-signed download URL for the attachment file
  - Should use the original `file_name` for the download
- [ ] Returns the attachment metadata
    - Excluding `object_key`
    - Including the download url

rowan04:
## Description

Implements a GET endpoint `/attachments/<attachment_id>` that:

- Generates a pre-signed download URL for the attachment file
   - Uses the original `file_name` for the download
- Returns the attachment metadata
   - Excludes `object_key`
   - Includes the download url
- Gracefully deals with a situation when the file object with the given `object_key` doesn't exist

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

Resolves #26


asuresh-code (changes requested):
Works well, only have a few minor suggestions.

object_storage_api/stores/attachment.py
```diff
@@ -16,6 +17,26 @@ class AttachmentStore:
...
     """
 
+    def create_presigned_get(self, attachment: AttachmentOut) -> str:
+        """
+        Generate a presigned url to share an S3 object.
+
+        :param attachment: `AttachmentOut` model of the attachment
+        :return: Presigned url to get the attachment.
+        """
+        logger.info("Generating presigned url to get attachment from object storage")
```
asuresh-code, line 27:
```suggestion
        logger.info("Generating presigned url to get attachment from object storage")
```
Might be worth including the object key in the logging message (similar to the image store message) to help with debugging in the future.

object_storage_api/stores/attachment.py
```diff
@@ -16,6 +17,26 @@ class AttachmentStore:
...
+        response = s3_client.generate_presigned_url(
+            "get_object",
+            Params={
+                "Bucket": object_storage_config.bucket_name.get_secret_value(),
+                "Key": attachment.object_key,
+                "ResponseContentDisposition": f'inline; filename="{attachment.file_name}"',
```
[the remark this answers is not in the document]
joelvdavies: In this case as we only need to download attachments and dont view them on the front end, I would recommend just using the attachment here.

asuresh-code (dismissed):

object_storage_api/schemas/attachment.py
```diff
@@ -21,27 +19,27 @@ class AttachmentPostSchema(BaseModel):
...
 
     upload_info: AttachmentPostUploadInfoSchema = Field(
         description="Information required to upload the attachment file"
     )
+
+
+class AttachmentSchema(AttachmentMetadataSchema):
+    """Schema model for an attachment get request response."""
+
+    url: HttpUrl = Field(description="Presigned get URL to get the attachment file")
```
joelvdavies, line 45:
```suggestion
    dowload_url: HttpUrl = Field(description="Presigned get URL to download the attachment file")
```
To match how we are now distinguishing URLs in #94.

test/e2e/test_attachment.py
```diff
@@ -136,11 +137,60 @@ def test_create_with_invalid_entity_id(self):
...
+        """
+        Checks that a prior call to `get_attachment` gave a successful response with the expected data returned.
+
+        :param expected_attachment_data: Dictionary containing the expected attachment data as would be required
+            for an `AttachmentMetadataSchema`.
+        """
+        assert self._get_response_attachment.status_code == 200
+        assert self._get_response_attachment.json() == expected_attachment_data
+
+    def check_get_attachment_failed(self) -> None:
```
joelvdavies, line 163:
```suggestion
    def check_get_attachment_failed(self, status_code: int, detail: str) -> None:
```
Could this match IMS and take in the message instead? I think this makes it a bit clearer what the type of failure is in the test itself and is easier to add more in future. Could you do the same to the corresponding image one as well please.

joelvdavies (commented):
Looks good, two minor comments.

VKTB (commented):
Considering @joelvdavies's comment here https://github.com/ral-facilities/object-storage-api/pull/91#discussion_r1918110706, my suggested changes are as follows:
- Rename `create_presigned_get` method in `stores/attachment.py` to `create_presigned_download_url`. Download URL to me implies that it would be a `GET`.
- Rename `create_presigned_get` in `stores/images.py` to something like `create_presigned_view_url`. View URL to me implies that it would be a `GET`.
- Create a new method called `create_presegined_download_url` in `stores/images.py` that will generate a presigned URL for downloading images. Download URL to me implies that it would be a `GET`.
- Rename `create_presigned_post` methods in `stores/attachment.py` and `stores/images.py` to `create_presigned_upload_url`. Upload URL to me implies that it would be a `POST`.

Let me know your thoughts @joelvdavies, @rowan04, @asuresh-code. This can be addressed separately.

joelvdavies:
> Considering @joelvdavies's comment here [#91 (comment)](https://github.com/ral-facilities/object-storage-api/pull/91#discussion_r1918110706), my suggested changes are as follows:
> 
> * Rename `create_presigned_get` method in `stores/attachment.py` to `create_presigned_download_url`. Download URL to me implies that it would be a `GET`.
> * Rename `create_presigned_get` in `stores/images.py` to something like `create_presigned_view_url`. View URL to me implies that it would be a `GET`.
> * Create a new method called `create_presegined_download_url` in `stores/images.py` that will generate a presigned URL for downloading images. Download URL to me implies that it would be a `GET`.
> * Rename `create_presigned_post` methods in `stores/attachment.py` and `stores/images.py` to `create_presigned_upload_url`. Upload URL to me implies that it would be a `POST`.
> 
> Let me know your thoughts @joelvdavies, @rowan04, @asuresh-code. This can be addressed separately.
Yes I don't mind between this or the current implementation, I think this is probably a bit clearer for our purposes. I wrote the comment above before I had seen what Anikesh would do in #93, which I think also works without changing anything as the `get` refers to it being as part of the get request and  just returns both urls in one (I don't think there is currently a requirement to split them). The original naming just came from boto3, which has both a create_presigned_url and create_presigned_post method, the latter being a special case which is why I originally named it after it.

rowan04:
> Yes I don't mind between this or the current implementation, I think this is probably a bit clearer for our purposes. I wrote the comment above before I had seen what Anikesh would do in #93, which I think also works without changing anything as the `get` refers to it being as part of the get request and just returns both urls in one (I don't think there is currently a requirement to split them). The original naming just came from boto3, which has both a create_presigned_url and create_presigned_post method, the latter being a special case which is why I originally named it after it.

So should I implement Viktor's suggested changes or not? And if so, in this PR/issue or a separate one?

VKTB:
> So should I implement Viktor's suggested changes or not? And if so, in this PR/issue or in a separate one?

Don't worry about addressing it as part of this PR, we can address it separately if we agree that we want to do this.

VKTB (approved):

joelvdavies (approved):

PR #94, opened 2025-01-16
Add download url to endpoint for getting a single image #93

### LINKED ISSUE #93: Add `download_url` to endpoint for getting a single image
This new url should use attachment instead of inline and be returned alongside the `url`. This will allow the front end to download them more easily.

asuresh-code:
## Description

See #93
Needs to wait for corresponding front end PR to be ready, before merging.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #93

### DIFF
--- a/object_storage_api/services/image.py
+++ b/object_storage_api/services/image.py
@@ -82,10 +82,10 @@ class ImageService:
 
     def get(self, image_id: str) -> ImageSchema:
         """
-        Retrieve an image's metadata with its presigned get url by its ID.
+        Retrieve an image's metadata with its presigned get download and inline urls by its ID.
 
         :param image_id: ID of the image to retrieve.
-        :return: An image's metadata with a presigned get url.
+        :return: An image's metadata with its presigned get urls.
         """
         image = self._image_repository.get(image_id=image_id)
         (inline_url, download_url) = self._image_store.create_presigned_get(image)
--- a/object_storage_api/services/image.py
+++ b/object_storage_api/services/image.py
@@ -88,7 +88,7 @@ class ImageService:
         :return: An image's metadata with its presigned get urls.
         """
         image = self._image_repository.get(image_id=image_id)
-        (view_url, download_url) = self._image_store.create_presigned_get(image)
+        view_url, download_url = self._image_store.create_presigned_get(image)
         return ImageSchema(**image.model_dump(), view_url=view_url, download_url=download_url)
 
     def list(self, entity_id: Optional[str] = None, primary: Optional[bool] = None) -> list[ImageMetadataSchema]:
--- a/object_storage_api/services/image.py
+++ b/object_storage_api/services/image.py
@@ -88,8 +88,8 @@ class ImageService:
         :return: An image's metadata with a presigned get url.
         """
         image = self._image_repository.get(image_id=image_id)
-        presigned_url = self._image_store.create_presigned_get(image)
-        return ImageSchema(**image.model_dump(), url=presigned_url)
+        (inline_url, download_url) = self._image_store.create_presigned_get(image)
+        return ImageSchema(**image.model_dump(), inline_url=inline_url, download_url=download_url)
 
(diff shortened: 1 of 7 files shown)


test/mock_data.py
```diff
@@ -170,5 +170,6 @@
 
 IMAGE_GET_DATA_ALL_VALUES = {
     **IMAGE_GET_METADATA_DATA_ALL_VALUES,
-    "url": ANY,
+    "inline_url": ANY,
+    "download_url": ANY,
```
joelvdavies, line 174: I think it should probably either be
```suggestion
    "url": ANY,
    "download_url": ANY,
```
(Maybe "view_url")
Or otherwise
```suggestion
    "inline_url": ANY,
    "attachment_url": ANY,
```
I personally prefer the former, probably with `view_url`?

asuresh-code: I agree I think download_url with view_url is better

object_storage_api/stores/image.py
```diff
@@ -39,25 +39,38 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
...
+            "Params": {
                 "Bucket": object_storage_config.bucket_name.get_secret_value(),
                 "Key": image.object_key,
                 "ResponseContentDisposition": f'inline; filename="{image.file_name}"',
             },
-            ExpiresIn=object_storage_config.presigned_url_expiry_seconds,
+            "ExpiresIn": object_storage_config.presigned_url_expiry_seconds,
+        }
+
+        view_response = s3_client.generate_presigned_url(**parameters)
```
VKTB, line 61: I think that it would be best if we stay consistent everywhere - `view_response` holds the URL so I think that it should be renamed to `view_url`. Same for `attachment_response`, it should be called `download_url`.

This is also exactly why I made these suggestions for the name of the methods here https://github.com/ral-facilities/object-storage-api/pull/91#pullrequestreview-2567056179. It make more sense from a consistency perspective to name the methods like I suggested. Another reason is the `GET` URL(s) this method generates are quite specific and can't generate any type of `GET` URL.

object_storage_api/stores/image.py
```diff
@@ -39,25 +39,38 @@ def upload(self, image_id: str, image_metadata: ImagePostMetadataSchema, upload_
...
+                **parameters,
+                "Params": {
+                    **parameters["Params"],
+                    "ResponseContentDisposition": f'attachment; filename="{image.file_name}"',
+                },
+            }
         )
 
-        return response
+        return (view_response, attachment_response)
```
VKTB, line 73: The parentheses are redundant.

object_storage_api/services/image.py
```diff
@@ -82,14 +82,14 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
...
+        Retrieve an image's metadata with its presigned get download and view urls by its ID.
 
         :param image_id: ID of the image to retrieve.
-        :return: An image's metadata with a presigned get url.
+        :return: An image's metadata with its presigned get urls.
         """
         image = self._image_repository.get(image_id=image_id)
-        presigned_url = self._image_store.create_presigned_get(image)
-        return ImageSchema(**image.model_dump(), url=presigned_url)
+        (view_url, download_url) = self._image_store.create_presigned_get(image)
```
VKTB, line 91: The parentheses are redundant.

object_storage_api/stores/image.py
asuresh-code: Yeah, I think renaming the response to url is a good idea. 

In regards to creating 2 store methods for getting urls, instead of the current 1, I lean slightly towards creating 2 separate methods, just for readability reasons. Should I make that second change in this PR, or is that a separate issue?

joelvdavies: I would just create a new issue and link Viktor's original comment in the description.

rowan04 (approved):
lgtm!

VKTB (approved):

joelvdavies (approved):

Issue #79, opened 2024-12-12
Move invalid entity_id error handling for list attachments get endpoint into repositories layer


rowan04:
As discussed in #73.
Like ims-api does, for consistency.

PR #97, opened 2025-01-17
Implement endpoint for deleting attachment

### LINKED ISSUE #28: Add an endpoint for deleting an attachment
Implement a DELETE endpoint `/attachments/<attachment_id>` that:

- [ ] Deletes the attachment metadata from the database
- [ ] Deletes the file object in the object store

VKTB:
## Description
This PR implement a `/attachments/<attachment_id>` `DELETE` endpoint that allows consumers to delete an attachment by its ID. It deletes the attachment metadata from the database and the file object in the object store.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #28


VKTB:
TODOs will be addressed when https://github.com/ral-facilities/object-storage-api/pull/91 gets merged.

test/unit/stores/test_attachment.py
```diff
@@ -30,6 +30,40 @@ def setup(self):
...
+# pylint: disable=duplicate-code
+
+
+class DeleteDSL(AttachmentStoreDSL):
+    """Base class for `delete` tests."""
+
+    _delete_object_key: str
+
+    def call_delete(self, object_key: str) -> None:
+        """Calls the `AttachmentStore` `delete` method."""
```
rowan04, line 43:
```suggestion
        """
        Calls the `AttachmentStore` `delete` method.

        :param object_key: Key of the attachment to delete.
        """
```

asuresh-code:
> I can delete an attachment fine. But if you try to then delete the same attachment_id again, it returns 500 "detail": "Something went wrong", and this in terminal. It doesn't do this for images, images returns a 404.

The problem here is the attachment get method (to checks if it exists) , does not raise an error like the image get request. The attachment get repo method would either have to be changed to match the images get repo method, or there needs to be a check for a None type object in the service layer before calling the subsequent delete methods.

asuresh-code (changes requested):
The implementation and testing looks good. Only changes would be what Rowan suggested and my reply to one of his comments.

rowan04:
> The problem here is the attachment get repo method (to checks if it exists) , does not raise an error like the image get repo method. The attachment get repo method would either have to be changed to match the images get repo method, or there needs to be a check for a None type object in the service layer before calling the subsequent delete methods.
> 
> Also, I think the raise `MissingRecordError` check in delete could be removed, since there is already a check earlier with the get repo method to see if doesn't exist. I think that check is in the Image version, because in the 1st draft of the PR an image was deleted from MongoDB and then the bucket, so that check was there instead of the get repo call, but then the order changed, requiring a get call, and I must've forgotten to remove that check (unless there is a reason we would want to keep it anyway)?

Ah yes - this was brought up in #73 and issue #79 was created for it

VKTB:
@rowan04 The issue you encountered will be resolved as soon as this branch is on top of your changes from https://github.com/ral-facilities/object-storage-api/pull/91.

> Also, I think the raise `MissingRecordError` check in delete could be removed, since there is already a check earlier with the get repo method to see if doesn't exist. I think that check is in the Image version, because in the 1st draft of the PR an image was deleted from MongoDB and then the bucket, so that check was there instead of the get repo call, but then the order changed, requiring a get call, and I must've forgotten to remove that check (unless there is a reason we would want to keep it anyway)?

I don't know what would be the right thing to do here. In IMS API repo layer we only raise `MissingRecordError` in the `delete` methods not the `get` methods.

test/e2e/test_attachment.py
```diff
@@ -254,3 +254,60 @@ def test_list_with_invalid_entity_id_filter(self):
...
+        Deletes an attachment with the given ID.
+        :param attachment_id: ID of the attachment to be deleted.
+        """
+        self._delete_response_attachment = self.test_client.delete(f"/attachments/{attachment_id}")
+
+    def check_delete_attachment_success(self) -> None:
+        """Checks that a prior call to `delete_attachment` gave a successful response with the expected code."""
+        assert self._delete_response_attachment.status_code == 204
+
+    def check_delete_attachment_failed_with_detail(self) -> None:
```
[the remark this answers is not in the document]
VKTB: I am guessing then that `check_get_attachment_failed` should be renamed to `check_get_attachment_failed_with_detail` and `check_get_image_failed` to `check_get_image_failed_with_detail`?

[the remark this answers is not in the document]
asuresh-code: I would say so, yes.

asuresh-code:
> @asuresh-code I don't know what would be the right thing to do here. In IMS API repo layer we only raise `MissingRecordError` in the `delete` methods not the `get` methods.

When I was working on this PR, @joelvdavies suggested looking into moving some of the error handling into the repo layer (like `MissingRecordError`) since it seemed more appropriate than what we were doing in IMS API. I would say probably follow what we did in the Image repo layer for now. 

I believe we are meant to have a discussion/separate issue to look into whether we want to adopt this approach more widely (moving error handling to the repo layer). I would say its probably best to create a separate issue to see if we want to revert these changes for both Images and Attachments.

joelvdavies:
> > @asuresh-code I don't know what would be the right thing to do here. In IMS API repo layer we only raise `MissingRecordError` in the `delete` methods not the `get` methods.
> 
> When I was working on this PR, @joelvdavies suggested looking into moving some of the error handling into the repo layer (like `MissingRecordError`) since it seemed more appropriate than what we were doing in IMS API. I would say probably follow what we did in the Image repo layer for now.
> 
> I believe we are meant to have a discussion/separate issue to look into whether we want to adopt this approach more widely (moving error handling to the repo layer). I would say its probably best to create a separate issue to see if we want to revert these changes for both Images and Attachments.

Yes its already in https://github.com/ral-facilities/object-storage-api/issues/15. The idea is that the errors happen where they occur then can be caught and changed/ignored otherwise but if forgotten about will propagate correctly all the way up to give a response to the request without having to replicate the logic in the router. The get is one area I was not 100% on considering the try and catch doesn't look as nice, its also kind of related to https://github.com/ral-facilities/inventory-management-system-api/issues/158.

object_storage_api/services/attachment.py
```diff
@@ -91,3 +91,13 @@ def list(self, entity_id: Optional[str] = None) -> list[AttachmentMetadataSchema
...
+
+    def delete(self, attachment_id: str) -> None:
+        """
+        Delete an attachment by its ID.
+        :param attachment_id: The ID of the attachment to delete.
+        """
+        stored_attachment = self._attachment_repository.get(attachment_id)
+        # Deletes attachment from object store first to prevent unreferenced objects in storage
+        self._attachment_store.delete(stored_attachment.object_key)
+        self._attachment_repository.delete(attachment_id)
```
asuresh-code, line 103:
```suggestion
    def delete(self, attachment_id: str) -> None:
        """
        Delete an attachment by its ID.
        :param attachment_id: The ID of the attachment to delete.
        """
        stored_attachment = self._attachment_repository.get(attachment_id)
        if stored_attachment is None:
           raise MissingRecordError(detail=f"No attachment found with ID: {attachment_id}", entity_name="attachment")
        # Deletes attachment from object store first to prevent unreferenced objects in storage
        self._attachment_store.delete(stored_attachment.object_key)
        self._attachment_repository.delete(attachment_id)
```
If no changes are made to the get repo layer for attachments, then I suggest adding a `NoneType` check in the service layer to see if it exists, and raising a `MissingRecordError` (404). Otherwise the 500 error Rowan pointed out will persist, when deleting a non-existent attachment.

asuresh-code (changes requested):

object_storage_api/services/attachment.py
VKTB: @asuresh-code That is no longer the case. Please refer to my reply to Rowan [here](https://github.com/ral-facilities/object-storage-api/pull/97#issuecomment-2609535908)

asuresh-code (approved):

rowan04 (approved):
lgtm!

joelvdavies (approved):

PR #105, opened 2025-01-29
Implement endpoint for deleting attachments by entity ID

### LINKED ISSUE #76: Add an endpoint for deleting attachments by entity ID
Dependent on #28. This endpoint should be a `DELETE` with a `entity_id` query parameter at `/attachments` that:
- [ ] Deletes all attachments with that `entity_id`
- [ ] Returns `204` if the `entity_id` is invalid or is not associated with any attachments

Will need consideration on how to deal with errors mid way, I expect they can be mitigated by deleting the database documents after successful deletion from the object store like we already do for the others.

VKTB:
## Description
This PR implements a `DELETE` endpoint at `/attachments` that accepts an `entity_id` query parameter and deletes all attachments with that `entity_id`.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Successfully deletes all attachments associated with the supplied `entity_id`
- [ ] Returns `204` if `entity_id` is invalid or not associated with any attachments

## Agile board tracking
closes #76

### DIFF
--- a/object_storage_api/routers/attachment.py
+++ b/object_storage_api/routers/attachment.py
@@ -8,6 +8,7 @@ from typing import Annotated, Optional
 
 from fastapi import APIRouter, Depends, Path, Query, status
 
+from object_storage_api.core.exceptions import InvalidObjectIdError
 from object_storage_api.schemas.attachment import (
     AttachmentMetadataSchema,
     AttachmentPostResponseSchema,
@@ -82,3 +83,23 @@ def delete_attachment(
     # pylint: disable=missing-function-docstring
     logger.info("Deleting attachment with ID: %s", attachment_id)
     attachment_service.delete(attachment_id)
+
+
+@router.delete(
+    path="",
+    summary="Delete attachments by entity ID",
+    response_description="Attachments deleted successfully",
+    status_code=status.HTTP_204_NO_CONTENT,
+)
+def delete_attachments_by_entity_id(
+    entity_id: Annotated[str, Query(description="The entity ID of the attachments to delete")],
+    attachment_service: AttachmentServiceDep,
+) -> None:
+    # pylint: disable=missing-function-docstring
+    logger.info("Deleting attachments with entity ID: %s", entity_id)
+    try:
+        attachment_service.delete_by_entity_id(entity_id)
+    except InvalidObjectIdError:
+        # As this endpoint takes in a query parameter to delete multiple attachments, and to hide the database
+        # behaviour, we treat any invalid entity_id the same as a valid one that has no attachments associated to it.
+        pass
--- a/object_storage_api/routers/attachment.py
+++ b/object_storage_api/routers/attachment.py
@@ -8,7 +8,6 @@ from typing import Annotated, Optional
 
 from fastapi import APIRouter, Depends, Path, Query, status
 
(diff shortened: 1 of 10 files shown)


object_storage_api/stores/attachment.py
```diff
@@ -88,3 +88,22 @@ def delete(self, object_key: str) -> None:
...
+        """
+        logger.info("Deleting attachment files with object keys: %s from the object store", object_keys)
+
+        batch_size = 1000
+        # Loop through the list of object keys in steps of `batch_size`
+        for i in range(0, len(object_keys), batch_size):
+            batch = object_keys[i : i + batch_size]
+            s3_client.delete_objects(
+                Bucket=object_storage_config.bucket_name.get_secret_value(),
+                Delete={"Objects": [{"Key": key for key in batch}]},
```
asuresh-code, line 108:
```suggestion
                Delete={"Objects": [{"Key": key } for key in batch]},
```
The current implementation works, but I'm not sure why it does?

The docs for the function requires the parameter to be structured as so:
image

If I modify the code to compare the 2 outputs of both options this is the result.
image
image

Again both implementation work, but I don't understand why yours works?

object_storage_api/routers/attachment.py
```diff
@@ -82,3 +83,23 @@ def delete_attachment(
...
+    attachment_service: AttachmentServiceDep,
+) -> None:
+    # pylint: disable=missing-function-docstring
+    logger.info("Deleting attachments with entity ID: %s", entity_id)
+    try:
+        attachment_service.delete_by_entity_id(entity_id)
+    except InvalidObjectIdError:
+        # As this endpoint takes in a query parameter to delete multiple attachments, and to hide the database
+        # behaviour, we treat any invalid entity_id the same as a valid one that has no attachments associated to it.
+        pass
```
asuresh-code, line 105: Whilst it is fine to catch the exception in the router layer, I think it would be better if it was in the repo layer to match what we are doing in the rest of the repo.

VKTB: Yes, we discussed this and I am going to move it to the repo layer.

asuresh-code (approved):
LGTM!

rowan04 (approved):
lgtm 🎉

test/unit/repositories/test_attachment.py
```diff
@@ -99,7 +99,7 @@ class GetDSL(AttachmentRepoDSL):
     _obtained_attachment_out: AttachmentOut
     _get_exception: pytest.ExceptionInfo
 
-    def mock_get(self, attachment_id: str, attachment_in_data: dict) -> None:
+    def mock_get(self, attachment_id: str, attachment_in_data: Optional[dict] = None) -> None:
```
joelvdavies, line 102:
```suggestion
    def mock_get(self, attachment_id: str, attachment_in_data: Optional[dict]) -> None:
```
Very minor point, but we always specify None in the mock gets everywhere else in IMS/for images here, so I would prefer to keep them consistent. I believe the idea is that its more readable by seeing the actual value in the test.

VKTB: I see. I will revert these changes then cheers.

joelvdavies (approved):

PR #119, opened 2025-02-17
Limit image file extensions that can be uploaded

### LINKED ISSUE #34: Limit file extension of images
The app should only support certain image file extensions. This should be configurable and the app should not allow images with extensions that are not supported to be uploaded.

VKTB:
## Description
This PR introduces a new config value (`IMAGE__ALLOWED_FILE_EXTENSIONS`) that can be used to configure the allowed file extensions for images that can be uploaded.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #34


asuresh-code (changes requested):
Works well, just have a few minor comments

object_storage_api/services/image.py
```diff
@@ -50,17 +51,21 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
...
 
-        expected_file_type = mimetypes.guess_type(upload_file.filename)[0]
+        expected_file_type, _ = mimetypes.guess_type(upload_file.filename)
         if expected_file_type != upload_file.content_type:
             raise InvalidFilenameExtension(
                 f"File extension `{upload_file.filename}` does not match content type `{upload_file.content_type}`"
             )
 
+        if expected_file_type not in config.image.allowed_mime_types:
+            raise InvalidFilenameExtension(f"File extension `{upload_file.filename}` is not supported")
```
asuresh-code, line 67:
```suggestion
        if upload_file.content_type not in config.image.allowed_mime_types:
            raise InvalidFilenameExtension(f"File type`{upload_file.content_type}` is not supported")
            
        if expected_file_type != upload_file.content_type:
            raise InvalidFilenameExtension(
                f"File extension `{upload_file.filename}` does not match content type `{upload_file.content_type}`"
            )
```
1. I think the check for whether it is an allowed file type should use the `upload_file.content_type` rather than the `expected_file_type`. Reason being is I think that the file content_type is more likely to be the actual image type and it is necessary for change 2.
2. I think the "allowed type" check should be before the "matching file extension" check. If someone tries to upload a file that has both a mismatched file extension, and invalid file type it makes more sense to tell them it is an unsupported file first, so that they don't try fixing the extension mismatch needlessly. 
**N.B: This would require change 1, since it otherwise could be a supported file type with a mismatched file extension, i.e an image of content type `image/jpeg` called `my_image.gif` would say unsupported file extension, when it's actually just a mismatched file extension.**

object_storage_api/.env.example
```diff
@@ -20,4 +20,5 @@ OBJECT_STORAGE__SECRET_ACCESS_KEY=example_password
 OBJECT_STORAGE__BUCKET_NAME=object-storage
 OBJECT_STORAGE__PRESIGNED_URL_EXPIRY_SECONDS=1800
 ATTACHMENT__MAX_SIZE_BYTES=104857600
+IMAGE__ALLOWED_MIME_TYPES=["image/jpeg", "image/png", "image/bmp", "image/tiff", "image/webp"]
```
asuresh-code, line 23:
```suggestion
IMAGE__ALLOWED_MIME_TYPES=["image/jpeg", "image/png", "image/bmp", "image/tiff", "image/webp", "image/heic", "image/heif"]
```
IPhones released after 2017 store images as a HEIC/HEIF file by default instead of JPEG. They do automatically convert to JPEG if you share them using some methods (like email), but stay as a HEIC/F if you upload them (like using Google Drive). 

If they're using their phones to take photos and upload them this could pose a problem. I think most android phones use JPEG as their default; I think some support HEIC, but I'm not sure if it's set as default or not.

VKTB: I can't see this on the [page](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html) about supported file formats by Pillow so I am assuming it will just break when it tries to open an image of this type?

joelvdavies: Good spot, I think I have an image somewhere I can try.

object_storage_api/services/image.py
joelvdavies: Just had a quick thought as well, I don't know how it would be limited on the front end whether its by file extension or content type, but I was thinking it would be nice if they are on the same thing so its effectively the same config setting?

VKTB:
We made the decision to to limit image extensions instead. The same will be implemented for attachments as part of https://github.com/ral-facilities/object-storage-api/issues/125.

object_storage_api/services/image.py
VKTB: See https://github.com/ral-facilities/object-storage-api/pull/119#issuecomment-2681765675

asuresh-code (changes requested):
Just have some suggestions regarding changing the allowed environment variable to file extensions instead of types.

object_storage_api/services/image.py
```diff
@@ -50,14 +52,19 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
...
         :raises InvalidObjectIdError: If the image has any invalid ID's in it.
-        :raises InvalidFilenameExtension: If the image has a mismatched file extension.
+        :raises InvalidFilenameExtension: If the image has a mismatched file extension or the file extension is not
+            supported.
         """
 
-        expected_file_type = mimetypes.guess_type(upload_file.filename)[0]
+        file_extension = Path(upload_file.filename).suffix
+        if not file_extension or file_extension not in config.image.allowed_file_extensions:
+            raise FileTypeMismatchException(f"File extension `{upload_file.filename}` is not supported")
```
asuresh-code, line 63:
```suggestion
        file_extension = Path(upload_file.filename).suffix
        if not file_extension or file_extension.toLowerCase() not in config.image.allowed_file_extensions:
            raise FileTypeMismatchException(f"File extension `{upload_file.filename}` is not supported")
```
File extensions can be capitalised sometimes, so you could either do an implementation like this or add the capitalised version of extensions to the allowed list

object_storage_api/.env.example
```diff
@@ -20,4 +20,5 @@ OBJECT_STORAGE__SECRET_ACCESS_KEY=example_password
 OBJECT_STORAGE__BUCKET_NAME=object-storage
 OBJECT_STORAGE__PRESIGNED_URL_EXPIRY_SECONDS=1800
 ATTACHMENT__MAX_SIZE_BYTES=104857600
+IMAGE__ALLOWED_FILE_EXTENSIONS=[".jpg", ".jpeg", ".png", ".bmp", ".tiff", ".webp"]
```
asuresh-code, line 23:
```suggestion
IMAGE__ALLOWED_FILE_EXTENSIONS=[".jpg", ".jpeg", ".jpe",".png", ".bmp", ".tiff", ".tif", ".webp"]
```
There are few more valid extensions for some of these content types. The mimetypes library has a content-type to file extension mapping that you can access to see all of the valid file extensions they cover (I've added all of the missing ones in my suggestion).

You can see the dictionary yourself with the code below.
```
mimetype = mimetypes.MimeTypes()
print(mimetype.types_map_inv)
```

VKTB: Thanks for that!

object_storage_api/services/image.py
```diff
@@ -50,14 +52,19 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
...
-        expected_file_type = mimetypes.guess_type(upload_file.filename)[0]
+        file_extension = Path(upload_file.filename).suffix
+        if not file_extension or file_extension.lower() not in config.image.allowed_file_extensions:
+            raise FileTypeMismatchException(f"File extension `{upload_file.filename}` is not supported")
+
+        expected_file_type, _ = mimetypes.guess_type(upload_file.filename)
         if expected_file_type != upload_file.content_type:
-            raise InvalidFilenameExtension(
+            raise FileTypeMismatchException(
                 f"File extension `{upload_file.filename}` does not match content type `{upload_file.content_type}`"
```
joelvdavies, line 68:
```suggestion
                f"File extension of '{upload_file.filename}' does not match content type `{upload_file.content_type}`"
```
I don't think this made sense before? E.g. `File extension `nvda_2024.4.2.exe` is not supported`

test/e2e/test_image.py
```diff
@@ -125,7 +125,7 @@ def test_create_with_file_extension_content_type_mismatch(self):
         self.post_image_with_file_extension_content_type_mismatch(
             IMAGE_POST_METADATA_DATA_REQUIRED_VALUES_ONLY, "image.jpg"
         )
-        self.check_post_image_failed_with_detail(422, "Filename does not contain the correct extension")
+        self.check_post_image_failed_with_detail(422, "File extension and content type do not match")
```
joelvdavies, line 128: I think there should be another e2e test for a file type that is known not to be allowed, rather than just mismatching.

joelvdavies (commented):
I noticed if I give it a file without an extension it gives an error like
`File extension `manual-test` is not supported` and responds with `File extension and content type do not match`, I think the latter should probably be the UnsupportedFileExtensionException if it the extension doesn't exist entirely?

VKTB:
> I noticed if I give it a file without an extension it gives an error like `File extension `manual-test` is not supported` and responds with `File extension and content type do not match`, I think the latter should probably be the UnsupportedFileExtensionException if it the extension doesn't exist entirely?

I was meaning to use `UnsupportedFileExtensionException`, good spot cheers!

object_storage_api/services/image.py
```diff
@@ -50,15 +56,20 @@ def create(self, image_metadata: ImagePostMetadataSchema, upload_file: UploadFil
 
         :param image_metadata: Metadata of the image to be created.
         :param upload_file: Upload file of the image to be created.
-        :return: Created image with an pre-signed upload URL.
+        :return: Created image with a pre-signed upload URL.
         :raises InvalidObjectIdError: If the image has any invalid ID's in it.
-        :raises InvalidFilenameExtension: If the image has a mismatched file extension.
+        :raises InvalidFilenameExtension: If the image has a mismatched file extension or the file extension is not
+            supported.
         """
```
asuresh-code, line 63: Comments need to be updated to reflect the exception name changes. Also needs to be changed in the update function in this file.

asuresh-code (changes requested):
Works well, just have a minor documentation suggestion.

object_storage_api/services/image.py
VKTB: Looks like I partially addressed this https://github.com/ral-facilities/object-storage-api/pull/133/files but I will fully address it here and resolve the conflicts in https://github.com/ral-facilities/object-storage-api/pull/133/files.

joelvdavies (approved):

test/e2e/test_image.py
```diff
@@ -125,7 +125,13 @@ def test_create_with_file_extension_content_type_mismatch(self):
...
+
+    def test_create_with_unsupported_file_extension(self):
+        """Test creating an image with an unsupported file extension."""
+
+        self.post_image(IMAGE_POST_METADATA_DATA_REQUIRED_VALUES_ONLY, "image.webp")
```
rowan04, line 133: why test with an extension that would be allowed according to `object_storage_api/.env.example`? Wouldn't it make more sense to use an extension that is completely not allowed (like .gif). Or is there a reason for doing this (like to test it works when a usually supported extension isn't exported in this env)?

rowan04 (commented):
Looks good! 1 comment

asuresh-code (approved):

test/e2e/test_image.py
VKTB: It's only an example so the actual file doesn't necessarily need to support it. Also depending on needs `.gif` might be allowed so. It was simpler and quicker at the time for me to just convert the existing `image.jpg` to `image.webp`.

I can look into ways of creating a `.gif` image and using that over `.webp` if others also think the same.

rowan04: Ok, well I'm happy for it to stay as `.webp` then, unless someone else thinks it should change.

rowan04 (approved):
lgtm!

PR #121, opened 2025-02-20
Add a configuration for limiting the number of attachments that can be uploaded

### LINKED ISSUE #111: Add a configuration for limiting the number of attachments that can be uploaded
A configuration should be added that will limit the number of attachments that can be uploaded.

VKTB:
## Description
This PR introduces a new config value (`ATTACHMENT__MAX_UPLOAD_LIMIT`) that can be used to configure the number of attachments that can be uploaded. If an attachment is created and the limit is reached, the API will not create it and return `422`. This is a soft limit which means there may be cases where it may be surpassed but not by a huge margin. This is fine for now but we are planning to improve this as part of https://github.com/ral-facilities/object-storage-api/issues/128.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #111


test/pytest.ini
```diff
@@ -23,4 +23,5 @@ env =
     OBJECT_STORAGE__BUCKET_NAME=test-object-storage
     OBJECT_STORAGE__PRESIGNED_URL_EXPIRY_SECONDS=1800
     ATTACHMENT__MAX_SIZE_BYTES=100
+    ATTACHMENT__UPLOAD_LIMIT=50
```
joelvdavies, line 26: Should this be lowered and have an e2e test?

joelvdavies (commented):
Work well for me even when uploading multiple, just one comment.

Also could you merge develop into here please, the docker-compose is outdated after the recent docker changes.

test/pytest.ini
joelvdavies: I don't think there are any cases we need more than one at the moment? So could probably even do 2 but I doubt it should matter much - if we need to raise it for others tests in future we can.

joelvdavies (approved):
LGTM

test/unit/repositories/test_attachment.py
```diff
@@ -544,3 +543,55 @@ def test_update_with_invalid_id(self):
...
+        self.check_count_by_entity_id_success()
+
+    def test_count_by_entity_id_with_no_results(self):
+        """Test counting all attachments returning no results."""
+        self.mock_count_by_entity_id(0)
+        self.call_count_by_entity_id(str(ObjectId()))
+        self.check_count_by_entity_id_success()
+
+
+# pylint: enable=duplicate-code
```
rowan04, line 597:
```suggestion
```
it never gets disabled after line 433 no?

rowan04 (commented):
looks good, one tiny comment

test/unit/repositories/test_attachment.py
rowan04: actually ive just seen the disable gets added in the image_limit configuration PR (#122), so I'll just resolve this.

rowan04 (approved):
lgtm!

object_storage_api/core/exceptions.py
```diff
@@ -61,6 +61,13 @@ class InvalidFilenameExtension(BaseAPIException):
     response_detail = "Filename does not contain the correct extension"
 
 
+class AttachmentUploadLimitReached(BaseAPIException):
+    """The limit for the maximum number of attachments for the provided `entity_id` has been reached."""
+
+    status_code = 422
+    response_detail = "Limit for the maximum number of attachments for the provided `entity_id` has been reached"
```
asuresh-code, line 68: Images also has a very similar exception. To reduce code duplication you could merge both exceptions into one, and use a parameter to specify if it's for images or attachments (similar to the `MissingRecordError`)

asuresh-code (changes requested):
Looks good, just have 1 minor change to suggest.

asuresh-code (approved):

joelvdavies (approved):

PR #122, opened 2025-02-21
Add a configuration for limiting the number of images that can be uploaded

### LINKED ISSUE #112: Add a configuration for limiting the maximum number of images that can be uploaded
A configuration should be added that will limit the maximum number of images that can be uploaded.

VKTB:
## Description
This PR introduces a new config value (`IMAGE__MAX_UPLOAD_LIMIT`) that can be used to configure the number of images that can be uploaded. If an image is created and the limit is reached, the API will not create it and return `422`. This is a soft limit which means there may be cases where it may be surpassed but not by a huge margin. This is fine for now but we are planning to improve this as part of https://github.com/ral-facilities/object-storage-api/issues/128.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #112


asuresh-code (changes requested):
Implementation works well for when you're uploading images 1 at a time, however when uploading images in batches on the front end using uppy, there are some issues.

When uploading multiple images using uppy, it seems to be processed in batches on the back end. For example, I had 98 images already uploaded for an entity, and uploaded 8 images. The first 5 seemed to be done as a batch; I put a log to output the current count of images, and there were 5 consecutive logs of `Current Count: 98` for each image. The remaining 3 images then all printed a value of  `Current Count: 103` images for the current count.

Currently, the check in the service layer uses a `==` comparison between the current count, and the max limit. You could change it to a `>=` comparison instead which would have prevented the last 3 images from uploading, however it would not have helped with the 1st batch of images, and would have resulted in 103 images being stored for an entity.

VKTB:
Following discussions, we decided for this to be a soft limit for the time being which means that there may be cases where it may be surpassed but not by a huge margin. We will change this to a hard limit as part of https://github.com/ral-facilities/object-storage-api/issues/128.

VKTB:
@asuresh-code Good spot!! You are right, the `>=` should have been used, thanks!

test/unit/services/test_image.py
```diff
@@ -139,19 +144,23 @@ def check_create_success(self) -> None:
...
         Checks that a prior call to `call_create_expecting_error` worked as expected, raising an exception
         with the correct message.
 
         :param message: Message of the raised exception.
+        :param assert_count: Whether the `count_by_entity_id` repo method is expected to be called or not.
         """
-        if assert_checks:
-            self.mock_generate_thumbnail_base64_str.assert_called_once_with(self._upload_file)
-            self.mock_image_store.upload.assert_called_once_with(
-                str(self._expected_image_id), self._image_post_metadata, self._upload_file
```
joelvdavies, line 152: Shouldn't these two still be here? Perhaps the variable could be `expect_count_check_success`, and if its None, it doesn't expect the check at all, and if its True/False it does expect it but only if its true does it assert these.

joelvdavies (commented):
Just the one comment, worked well. We could also limit uppy to one upload at a time to avoid going over the limit via the front end if we wanted to, Joshua and I originally agreed to let it do 5 in parallel which is what causes it. (Although would still be a soft limit otherwise)

test/unit/services/test_image.py
asuresh-code: I think it's fine if these set of checks are removed now. 

Before Viktor's Changes, the `invalid entity_id error` would be thrown after the thumnbail is generated, and the image is uploaded (requiring the check, since some failures would throw exceptions before these operations). After Viktor's changes, the entity_id check has been moved to the start of the service method, so the checks are no longer necessary.

asuresh-code (changes requested):
Works Well, testing using the endpoint in the documents and the frontend UI. Just have 1 minor documentation request change, besides that it all looks good.

test/unit/services/test_image.py
VKTB: `None` can be assigned to a `bool` variable. But yes, the reason I removed them is basically what Anikesh outlined.

asuresh-code (approved):
LGTM

test/e2e/test_image.py
```diff
@@ -69,6 +69,42 @@ def post_image_with_file_extension_content_type_mismatch(
...
+        )
+
+        # Third image
+        image_c_id = self.post_image(
+            {
+                **IMAGE_POST_METADATA_DATA_ALL_VALUES,
+                "entity_id": entity_id_b,
+            },
+            "image.jpg",
+        )
```
rowan04, line 100:
```suggestion
        # Second image
        image_b_id = self.post_image({**IMAGE_POST_METADATA_DATA_ALL_VALUES, "entity_id": entity_id_a}, "image.jpg")

        # Third image
        image_c_id = self.post_image({**IMAGE_POST_METADATA_DATA_ALL_VALUES, "entity_id": entity_id_b}, "image.jpg")
```
to match `image_a_id`

test/unit/services/test_image.py
```diff
@@ -139,19 +144,23 @@ def check_create_success(self) -> None:
...
+        if assert_count:
+            self.mock_image_repository.count_by_entity_id.assert_called_once_with(
+                str(self._expected_image_in.entity_id)
             )
-        self.mock_image_repository.create.assert_not_called()
+        else:
+            self.mock_image_repository.count_by_entity_id.create.assert_not_called()
+            self.mock_generate_thumbnail_base64_str.create.assert_not_called()
+            self.mock_image_store.upload.create.assert_not_called()
+            self.mock_image_repository.create.assert_not_called()
```
rowan04, line 164:
```suggestion
        else:
            self.mock_image_repository.count_by_entity_id.create.assert_not_called()

        self.mock_generate_thumbnail_base64_str.create.assert_not_called()
        self.mock_image_store.upload.create.assert_not_called()
        self.mock_image_repository.create.assert_not_called()

```
should these be in the else block?

rowan04 (commented):
looks good overall! a few small comments

object_storage_api/core/exceptions.py
```diff
@@ -61,11 +61,16 @@ class InvalidFilenameExtension(BaseAPIException):
...
-class AttachmentUploadLimitReached(BaseAPIException):
-    """The limit for the maximum number of attachments for the provided `entity_id` has been reached."""
+class UploadLimitReachedError(BaseAPIException):
+    """The limit for the maximum number of attachments or images for the provided `entity_id` has been reached."""
 
     status_code = 422
-    response_detail = "Limit for the maximum number of attachments for the provided `entity_id` has been reached"
+
+    def __init__(self, detail: str, entity_name: str):
+        response_detail = (
```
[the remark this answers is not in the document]
VKTB: Yes, it should have. I completely forgot about it. Thanks!

joelvdavies (approved):

asuresh-code (dismissed):

joelvdavies (approved):

PR #133, opened 2025-03-12
Limit attachment file extensions #125

### LINKED ISSUE #125: Limit file extension of attachments
The app should only support certain attachment file extensions. This should be configurable and the app should not allow attachments with extensions that are not supported to be created.

VKTB:
## Description
This PR introduces a new config value (`ATTACHMENT__ALLOWED_FILE_EXTENSIONS`) that can be used to configure the allowed file extensions for attachments that can be uploaded.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #125


object_storage_api/.env.example
```diff
@@ -19,6 +19,7 @@ OBJECT_STORAGE__ACCESS_KEY=root
 OBJECT_STORAGE__SECRET_ACCESS_KEY=example_password
 OBJECT_STORAGE__BUCKET_NAME=object-storage
 OBJECT_STORAGE__PRESIGNED_URL_EXPIRY_SECONDS=1800
+ATTACHMENT__ALLOWED_FILE_EXTENSIONS=[".doc", ".pdf", ".txt"]
```
VKTB, line 22:
```suggestion
ATTACHMENT__ALLOWED_FILE_EXTENSIONS=[".doc", ".docx", ".pdf", ".txt"]
```

object_storage_api/.env.example
```diff
@@ -19,6 +19,7 @@ OBJECT_STORAGE__ACCESS_KEY=root
 OBJECT_STORAGE__SECRET_ACCESS_KEY=example_password
 OBJECT_STORAGE__BUCKET_NAME=object-storage
 OBJECT_STORAGE__PRESIGNED_URL_EXPIRY_SECONDS=1800
+ATTACHMENT__ALLOWED_FILE_EXTENSIONS=[".doc", ".docx", ".pdf", ".txt"]
```
joelvdavies, line 22: From the meeting with CLF they suggested pdfs, excel spreadsheets and word documents. So looking at https://learn.microsoft.com/en-us/office/compatibility/office-file-format-reference I would add `.csv`, `.xls` and `.xlsx`. Maybe the more open formats `.odt`, `.rtf`, `.ods` as well? Then again I doubt they would use those so we can just add them if needed.

joelvdavies (commented):
One comment, worked well for me.

object_storage_api/.env.example
rowan04: one of the example attachments i added in https://github.com/ral-facilities/inventory-management-system/pull/1296 had a `.rtf` extension.

test/pytest.ini
```diff
@@ -22,6 +22,7 @@ env =
     OBJECT_STORAGE__SECRET_ACCESS_KEY=example_password
     OBJECT_STORAGE__BUCKET_NAME=test-object-storage
     OBJECT_STORAGE__PRESIGNED_URL_EXPIRY_SECONDS=1800
+    ATTACHMENT__ALLOWED_FILE_EXTENSIONS=[".doc", ".pdf", ".txt"]
```
rowan04, line 25:
```suggestion
    ATTACHMENT__ALLOWED_FILE_EXTENSIONS=[".doc", ".docx", ".pdf", ".txt"]
```
why not `.docx`?

rowan04 (commented):
Looks good! 1 comment

asuresh-code (commented):
Works well, tested it using endpoints and the front end. If there is a batch mixed with supported and unsupported file types on the front end, only the supported ones are uploaded which is good.

The only change I'd recommend is including more allowed file types, specifically Joel's comment on spreadsheets, and Rowan's on docx. I don't think we would have to include anything outside of the [microsoft office suite's support file types](https://learn.microsoft.com/en-us/office/compatibility/office-file-format-reference).

object_storage_api/.env.example
VKTB: I can add this, sure. Though, this is only an example file and the actual file will always be different depending on the facility.

joelvdavies (dismissed):

asuresh-code (approved):

joelvdavies (approved):
Still works, changes look good to me.

PR #146, opened 2025-04-22
Upgrade boto3 and minio #143

### LINKED ISSUE #143: Resolve boto3/botocore upgrade e2e test failure
See https://github.com/ral-facilities/object-storage-api/pull/141. I had to downgrade the updated version of boto3/botocore in this PR in order to get the e2e tests to pass. We should figure out why this is the case and if possible fix it so we can update these too.

joelvdavies:
## Description

See #143. This updates boto3 to the latest version (at the time of writing) and also updates minio to fix the e2e tests that failed in the dependabot PRs. I also tested the new package version against echo to ensure it still functions correctly. docker-ims will need the same minio change in order for delete to work correctly.

Due to it being close to the demo I will not merge this or attempt update the demo VM just in case.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #143

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,7 @@ services:
       MONGO_INITDB_ROOT_PASSWORD: example
 
   minio:
-    image: minio/minio:RELEASE.2024-09-13T20-26-02Z
+    image: minio/minio:RELEASE.2025-04-08T15-41-24Z
     container_name: object-storage-minio
     command: minio server /data
     volumes:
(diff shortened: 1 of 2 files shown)


VKTB (approved):

VKTB:
If you could please open an issue in docker-ims so we won't forget, cheers.

joelvdavies:
> If you could please open an issue in docker-ims so we won't forget, cheers.

Done, you can merge this in whenever you are ready to, if its merged without the docker-ims change the only effect is that deleting wont work when using minio.

PR #151, opened 2025-04-30
Prevent images and attachments with duplicate file names #148

### LINKED ISSUE #148: Prevent images and attachments with duplicate names
Requested by Maks at the user feedback session. Main reason is to avoid potential for users naming something like `Current image` repeatedly for example and instead forcing the 2nd instance to have a different name.

joelvdavies:
## Description

See #148. Uses the same method as inventory-management-system-api uses for catalogue categories. As part of this the order of image upload and thumbnail generation has been moved to after the database insertion. This was due to a combination of reasons primarily due to wanting to have the duplicate check before so it can fail faster, but also wanting to avoid the possibility of a small image slipping in between a larger image upload after the larger images check has completed which would allow duplicate file names again. This issue is effectively the same as noted in https://github.com/ral-facilities/object-storage-api/pull/122#pullrequestreview-2657248393. By moving the count check after thumbnail generation too, the limit will be as hard as the attachment one is reducing the issue.

It is unfortunately still technically possible for it to allow the update (although I think incredibly unlikely) just due to the fact the database insert/edit is still separate to the actual check. I looked into used a database index for this, and found it should work well but is left as a modification after this work (I think the bulk write on images would need checking, Viktor suggested doing an UpdateOne just for the file data excluding the primary value, then update the primary value after the UpdateMany in another UpdateOne, meaning just the primary update would fail for example).

Overall changes
- Prevents creation/edit that would cause an attachment/image file name to be duplicated
- Moves the image upload to after the database insert during creation
- Moves the image/attachment count check to after the thumbnail generation during creation (reducing the impact of https://github.com/ral-facilities/object-storage-api/pull/122#pullrequestreview-265724839) allowing the image limit to be a harder limit again
- Renames `entity_name` to `entity_type` in the exceptions as I thought it could be confusing as its not really the name of the individual entity

The front end will likely need some modification to handle the new errors.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #148


object_storage_api/services/attachment.py
```diff
@@ -86,7 +80,21 @@ def create(self, attachment: AttachmentPostSchema) -> AttachmentPostResponseSche
...
-        attachment_in = AttachmentIn(**attachment.model_dump(), id=attachment_id, object_key=object_key)
+        attachment_in = AttachmentIn(
+            **attachment.model_dump(),
+            id=attachment_id,
+            code=utils.generate_code(attachment.file_name, "attachment"),
+            object_key=object_key,
+        )
+
+        # This should be checked as close to the actual create as possible to avoid concurrency issues
+        if self._attachment_repository.count_by_entity_id(attachment.entity_id) >= config.attachment.upload_limit:
```
asuresh-code, line 91: Out of curiosity, is there a reason this shouldn't be moved to the repo layer to ensure that it is as close to the actual create action?

It makes sense to keep it on the service layer semantically, since this is more related to business logic, rather than actual database interactions; however, in this case to address concurrency issues, would it make sense to move to the repo layer?

object_storage_api/repositories/attachment.py
```diff
@@ -111,20 +115,27 @@ def update(self, attachment_id: str, attachment: AttachmentIn, session: ClientSe
...
         except InvalidObjectIdError as exc:
             exc.status_code = 404
             exc.response_detail = "Attachment not found"
             raise exc
 
+        stored_attachment = self.get(str(attachment_id), session=session)
+        if attachment.file_name != stored_attachment.file_name and self._is_duplicate(
+            attachment.entity_id, attachment.code, attachment_id, session=session
+        ):
+            raise DuplicateRecordError("Duplicate attachment found within the parent entity", entity_type="attachment")
```
asuresh-code, line 132: What was the reason to include the 1st conditional on the file_name? The only one I can think of, is as you mentioned before, there is a chance duplicate names can end up in the system; in that case I assume you wouldn't want to trigger a duplicate error if they are simply updating the metadata of a file, which this file name condition would prevent.

asuresh-code (commented):
Just reviewed the implementation first, will look at testing later.

The database index does seem to fix the issue; I think you can create compound indexes and apply a uniqueness constraint on them, to ensure the database prevents the insert itself?

joelvdavies:
> The database index does seem to fix the issue; I think you can create compound indexes and apply a uniqueness constraint on them, to ensure the database prevents the insert itself?

Good point. This morning I was just adding the entity id to the code, but yeah looks like a compound index is a better idea, I didn't realise they existed 😄, thanks!

object_storage_api/repositories/attachment.py
joelvdavies: I didn't think of that, it was here just avoid doing the duplicate check (and database call) if the name hasn't changed at all as there isn't any point, it was copied from ims-api.

asuresh-code (dismissed):

VKTB (approved):

Issue #152, opened 2025-05-02
Add unique index to code field on attachments and images


joelvdavies:
Following on from #148, we should replace the manual check with a unique index on the database that prevents the duplicates - avoiding concurrency issues as much as possible.

- [ ] Check with database services we can do this ourself
- [ ] For development it would be nice if it was automatic like replica sets on ims-api
- [ ] Consider how this will be done in production (e.g. a script that can be executed/manual instructions)

PR #154, opened 2025-05-06
Use MongoDB indexing to prevent duplicate names #151

joelvdavies:
## Description

See #152. Ensures we shouldn't have race conditions due to concurrency which would otherwise allow some images/attachments with duplicate names from getting added to the database. This should not require any specific changes to your local instance, calling `docker compose up` should be sufficient to get the index to be setup.

Will require similar changes for docker-ims to handle the entrypoint script.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Ensure you cannot create images/attachments with duplicate file names (either though creating or editing)

## Agile board tracking

Closes #151

### DIFF
--- a/scripts/mongodb_entrypoint.sh
+++ b/scripts/mongodb_entrypoint.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# Execute the standard entrypoint to start mongodb, but add the additional parameters to setup the replica set
+echo "Starting MongoDB through default entrypoint..."
+
+# Avoid unbound variable error as there are no parameters in the initialisation command
+if [ "$#" -eq 0 ]; then
+  set -- mongod
+fi
+
+/usr/local/bin/docker-entrypoint.sh "$@" &
+
+# Wait for MongoDB to be ready to run commands on
+echo "Waiting for MongoDB to be ready..."
+until mongosh --username "$MONGO_INITDB_ROOT_USERNAME" --password "$MONGO_INITDB_ROOT_PASSWORD" --authenticationDatabase=admin --eval "db.adminCommand('ping')" > /dev/null 2>&1; do
+  sleep 1
+done
+
+# Initialise indexes if not already setup
+for database_name in object-storage test-object-storage; do
+  if ! mongosh "$database_name" --username "$MONGO_INITDB_ROOT_USERNAME" --password "$MONGO_INITDB_ROOT_PASSWORD" --authenticationDatabase=admin --quiet --eval "db.attachments.getIndexes().forEach(i => print(i.name))" | grep -q '^attachments_code_index$'; then
+    echo "Initialising code index for attachments and images ($database_name)..."
+    mongosh "$database_name" --username "$MONGO_INITDB_ROOT_USERNAME" --password "$MONGO_INITDB_ROOT_PASSWORD" --authenticationDatabase=admin \
+      --eval 'db.attachments.createIndex({ "code": 1 }, { name: "attachments_code_index", unique: true })' \
+      --eval 'db.images.createIndex({ "code": 1 }, { name: "images_code_index", unique: true })'
+  fi
+done
+
+wait
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,8 +23,10 @@ services:
   mongo-db:
     image: mongo:7.0-jammy
     container_name: object-storage-api-mongodb
+    entrypoint: mongodb_entrypoint.sh
     volumes:
       - ./mongodb/data:/data/db
+      - ./scripts/mongodb_entrypoint.sh:/usr/local/bin/mongodb_entrypoint.sh:ro
     restart: always
     ports:
       - 27018:27017


test/e2e/test_image.py
```diff
@@ -448,6 +456,37 @@ def test_update_file_name_to_duplicate(self):
             409, "An image with the same file name already exists within the parent entity."
         )
 
+    def test_update_file_name_to_duplicate_while_setting_primary(self):
+        """Test updating the name of an image to conflict with a pre-existing one while at the same time attempting to
+        set it as a primary (no image content should change)."""
```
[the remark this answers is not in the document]
joelvdavies: Yes your right it shouldn't, however we always duplicate checks on the backend and test them because it is still possible to use the api directly with the right knowledge.

joelvdavies:
> The implementation upon inspection looks good as well as the testing mostly.
> 
> I used `docker compose` to run the `object-storage-api`, and preventing duplicate attachments worked, however for images it did not. Since I used docker, I don't think I had to make any other changes?

Interesting, I think Joshua had a similar issue except neither worked as somehow his ports got confused or something so it actually added  the index to the ims database instead. You can look it up in MongoDB compass e.g.
image

In this case I would suggest running the following while its running
```bash
docker exec -i object-storage-api-mongodb mongosh object-storage --username root --password example --authenticationDatabase=admin --eval 'db.attachments.dropIndex("attachments_file_name_uniqueness_index")' --eval 'db.images.dropIndex("images_file_name_uniqueness_index")'
```
To ensure they are both removed, then do
```bash
docker compose down
docker compose up
```

and try it again.

README.md
```diff
@@ -200,6 +200,33 @@ Instances of these can be started using the `docker-compose.yml` file.
...
+##### MongoDB
+
+You must have access to a MongoDB instance with a database and be able to create the following indexes using either
+MongoDB Compass or a CLI e.g.
+
+```bash
+mongosh DATABASE_NAME --username USERNAME --password PASSWORD --authenticationDatabase=admin \
+   --eval 'db.attachments.createIndex({ "entity_id": 1, "code": 1 }, { name: "attachments_file_name_uniqueness_index", unique: true })' \
+   --eval 'db.images.createIndex({ "entity_id": 1, "code": 1 }, { name: "images_file_name_uniqueness_index", unique: true })'
+```
```
VKTB, line 214: Will these commands work if the collections have not been created? They only get created when the application starts and it adds something to the db.

joelvdavies: I tested it by dumping my database (and for good measure `sudo rm -rf mongodb/data/`), then running `docker compose up`, and it just creates the empty collections with the index. So it should work with or without any existing data (just as long as it doesn't already have the index)

asuresh-code:
I was able to fix it after inspecting the docker logs. I had duplicate images uploaded in my instance already, which was preventing the unique compound index from getting initialised.

It now seems to be behaving as expected on both the endpoint and the frontend.

joelvdavies:
> I was able to fix it after inspecting the docker logs. I had duplicate images uploaded in my instance already, which was preventing the unique compound index from getting initialised.
> 
> It now seems to be behaving as expected on both the endpoint and the frontend.

Ah I see, that would explain it, I didn't think of that. I also forgot about the mock data script, I found it errored at one point for me so I pushed a commit to copy the duplicate logic from ims-api.

scripts/mongodb_entrypoint.sh
```diff
@@ -0,0 +1,29 @@
...
+for database_name in object-storage test-object-storage; do
+  if ! mongosh "$database_name" --username "$MONGO_INITDB_ROOT_USERNAME" --password "$MONGO_INITDB_ROOT_PASSWORD" --authenticationDatabase=admin --quiet --eval "db.attachments.getIndexes().forEach(i => print(i.name))" | grep -q '^attachments_file_name_uniqueness_index$'; then
+    echo "Initialising unique file name index for attachments and images collections ($database_name)..."
+    mongosh "$database_name" --username "$MONGO_INITDB_ROOT_USERNAME" --password "$MONGO_INITDB_ROOT_PASSWORD" --authenticationDatabase=admin \
+      --eval 'db.attachments.createIndex({ "entity_id": 1, "code": 1 }, { name: "attachments_file_name_uniqueness_index", unique: true })' \
+      --eval 'db.images.createIndex({ "entity_id": 1, "code": 1 }, { name: "images_file_name_uniqueness_index", unique: true })'
+  fi
+done
+
+wait
```
VKTB, line 29:
```suggestion
wait

```

object_storage_api/repositories/image.py
```diff
@@ -130,23 +131,27 @@ def update(self, image_id: str, image: ImageIn, update_primary: bool, session: C
...
-            self._images_collection.bulk_write(bulkwrite_update, session=session)
-        else:
-            self._images_collection.update_one(
-                {"_id": image_id}, {"$set": image.model_dump(by_alias=True)}, session=session
-            )
+        try:
+            if update_primary:
+                bulkwrite_update = [
+                    UpdateOne(
+                        filter={"_id": image_id}, update={"$set": image.model_dump(by_alias=True, exclude="primary")}
```
VKTB, line 138: Will updates to other fields be made if the new name is found to be a duplicate?

VKTB, line 138: `Expected type 'set[int] | set[str] | Mapping[int, Literal[True] | Any] | Mapping[str, Literal[True] | Any] | None', got 'Literal["primary"]' instead`
```suggestion
                    UpdateOne(
                        filter={"_id": image_id}, update={"$set": image.model_dump(by_alias=True, exclude={"primary"})}
```

scripts/mongodb_entrypoint.sh
joelvdavies: Have just updated it. Is this just a convention for scripts? I seem to remember seeing something like it before.

object_storage_api/repositories/image.py
joelvdavies: I believe this is what is means for the operations being atomic (first paragraph of https://www.mongodb.com/docs/manual/core/write-operations-atomicity/#atomicity-and-transactions)

scripts/mongodb_entrypoint.sh
VKTB: From what I remember, there are standards for unix based systems that require each line in a text file to end with a new line (`\n`). I think GitHub warns about this for all files. I guess

object_storage_api/repositories/image.py
VKTB: Yes for the `UpdateOne`. In that case is there any point in excluding the `primary` field? If it fails due to a duplicate then `primary` will not get updated, and the `UpdateMany` will not run too.

joelvdavies: Technically we can leave it in, but then it will set it to True, then False then True again (For the UpdateOne, UpdateMany, UpdateOne), this current way avoids touching it so if it started false and is being changed to true, it only gets modified for that entity once at the end. We cant combine the first and last UpdateOne as then the UpdateMany would reset it to false.

VKTB: Yeah what you have done is exactly what I was trying to suggest. Cheers!

VKTB (dismissed):

VKTB (approved):

PR #167, opened 2025-05-22
Install package using pyproject.toml file in prod stage

joelvdavies:
## Description

Same as in https://github.com/ral-facilities/ldap-jwt-auth/pull/227.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


joelvdavies:
Just realised the change is more involved as it has a different structure currently, so closing to be addressed separately.

PR #283, opened 2026-03-27
Release v1.3.0

joelvdavies:
## Description

Merges `release/v1.3.0` into `main` in preparation for the `v1.3.0` release. Technically this does have a breaking change as the database env variables have been renamed. But it feels overkill to release as v2 as the rest are readme/package updates. As we use 1.2 in the prod config, releasing as 1.3 will still ensure it doesnt break any existing deployment, so I felt this would be alright here.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


joelvdavies:
Have tested this locally and on the dev machine using latest prod config.

